PR #7, opened 2024-08-07
feat: including typeORM on persistence-example template

melisaguevara:
Modifying the persistence-example to use TypeORM in that template.
Including scripts to generate and run migrations.


docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: "3.8"
-
```
melisaguevara, line 2: It seems like this was deprecated https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-obsolete

package.json
```diff
@@ -11,11 +11,15 @@
...
     "prepare": "husky",
+    "db:persistence-example:migrate:generate": "cd packages/persistence-example; pnpm db:migration:generate $MIGRATION_PATH",
+    "db:persistence-example:migrate:run": "cd packages/persistence-example; pnpm db:migration:run",
     "dev-env:up": "docker-compose up -d --build",
     "dev-env:down": "docker-compose down",
     "dev-env:run-app:indexer": "docker-compose exec app sh -c \"pnpm start:indexer\"",
     "dev-env:run-app:template": "docker-compose exec app sh -c \"pnpm start:template\"",
-    "dev-env:run-app:persistence-example": "docker-compose exec app sh -c \"pnpm start:persistence-example\""
+    "dev-env:run-app:persistence-example": "docker-compose exec app sh -c \"pnpm db:persistence-example:migrate:run && pnpm start:persistence-example\"",
+    "dev-env:persistence-example:db:migrate:generate": "docker-compose exec app sh -c \"pnpm db:persistence-example:migrate:generate $MIGRATION_PATH\"",
```
melisaguevara, line 21: Had to include $MIGRATION_PATH because the command for running migrations needs the path to the migration folder.

Looked for a way to have the script be for example:
` "db:migration:generate": "pnpm typeorm migration:generate -d migrations.config.ts ./path/to/migrations/<migration-name>",
`
To be able to run it doing:
`pnpm run db:migration:generate <migration-name>`
but I couldn't find it.  

lmk if u guys know any way to accomplish this.

james-a-morris: I'm not familiar with a better way to do this

packages/typescript-config/base.json
```diff
@@ -15,6 +17,7 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "ES2022"
+    "strictPropertyInitialization": false,
```
james-a-morris, line 20: qq: what was the reason for this addition?

james-a-morris (approved):
lgtm - one comment

packages/typescript-config/base.json
melisaguevara: This is needed for typeorm [entity files](packages/persistence-example/src/entities/User.entity.ts). Otherwise compilation will fail with the error "Property 'id' has no initializer and is not definitely assigned in the constructor."

package.json
daywiss: im sure you could do it with a bash script but dont think its worth teh complexity

packages/persistence-example/src/data-source.ts
```diff
@@ -0,0 +1,20 @@
+import "reflect-metadata";
+import { DataSource } from "typeorm";
+import { User } from "./entities/User.entity";
+
+export const createDataSource = (
+  env: Record<string, string | undefined>,
```
daywiss, line 6: imo its probably better to strictly define the config you expect here and do env parsing at the caller. this will give us a convention where we can quickly locate what envs are expected and where they are parsed. also makes this file a bit more composable if used elsewhere ( not that its expected to be )

package.json
```diff
@@ -11,11 +11,15 @@
...
     "start:persistence-example": "APP=persistence-example pnpm --filter @repo/node-app start",
     "prepare": "husky",
+    "db:persistence-example:migrate:generate": "cd packages/persistence-example; pnpm db:migration:generate $MIGRATION_PATH",
+    "db:persistence-example:migrate:run": "cd packages/persistence-example; pnpm db:migration:run",
     "dev-env:up": "docker-compose up -d --build",
     "dev-env:down": "docker-compose down",
     "dev-env:run-app:indexer": "docker-compose exec app sh -c \"pnpm start:indexer\"",
     "dev-env:run-app:template": "docker-compose exec app sh -c \"pnpm start:template\"",
-    "dev-env:run-app:persistence-example": "docker-compose exec app sh -c \"pnpm start:persistence-example\""
+    "dev-env:run-app:persistence-example": "docker-compose exec app sh -c \"pnpm db:persistence-example:migrate:run && pnpm start:persistence-example\"",
```
daywiss, line 20: btw shoudl have mentioned this before, but isnt it standard now to use `docker compose`, rather than `docker-compose`

daywiss (approved):

packages/typescript-config/base.json
james-a-morris: Good call 🫡

package.json
james-a-morris: Agreed - it's a shame that PNPM doesn't set any ENV vars to indicate that the code is being called from a script

package.json
james-a-morris: I didn't realize these were the same binary now - good call +1

melisaguevara: Yep it seems like it is https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2

melisaguevara: I've updated it in this commit [59f7510](https://github.com/across-protocol/indexer/pull/7/commits/59f7510719c72074fa78a1e075f8ff8f22440f0c)

packages/persistence-example/src/data-source.ts
melisaguevara: I agree yes, that makes sense
Something like this for example? [59f7510](https://github.com/across-protocol/indexer/pull/7/commits/59f7510719c72074fa78a1e075f8ff8f22440f0c)

daywiss (approved):
great

PR #8, opened 2024-08-08
feat: set up indexer-database package

melisaguevara:


packages/indexer-api/src/database/database.provider.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function connectToDatabase(
+  env: Record<string, string | undefined>,
+) {
+  const databaseConfig = {
+    host: env.DATABASE_HOST || "localhost",
+    port: env.DATABASE_PORT || "5432",
+    user: env.DATABASE_USER || "user",
+    password: env.DATABASE_PASSWORD || "password",
+    dbName: env.DATABASE_NAME || "database",
```
james-a-morris, line 11: I think we should enforce these variables and not have these defaults

packages/indexer-database/README.md
```diff
@@ -0,0 +1,79 @@
...
+#### Running migrations
+To execute migrations, run:
+```
+pnpm run db:migration:run
+```
+
+### Using this package
+1. Add it as a dependency of other packages that need to interact with the indexer database:
+```json
+// other package package.json
```
james-a-morris, line 49: Nit: JSON (not json5) doesn't have comments embedded into the string - I think github's diff is throwing an error on this

melisaguevara: Ty! I deleted the extension to be able to keep the comments

packages/indexer-api/src/database/database.provider.ts
melisaguevara: Changed here [502f045](https://github.com/across-protocol/indexer/pull/8/commits/502f04592c7b0b76ce1169a0257146f962944e2f)

packages/indexer-database/src/main.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const createDataSource = (config: {
+  host: string | undefined;
+  port: string | undefined;
+  user: string | undefined;
+  password: string | undefined;
+  dbName: string | undefined;
+}): DataSource => {
+  for (const [key, value] of Object.entries(config)) {
+    console.log(key, value);
```
daywiss, line 12: stray log here

packages/indexer-database/src/main.ts
```diff
@@ -0,0 +1,28 @@
...
+import { DataSource } from "typeorm";
+
+export const createDataSource = (config: {
+  host: string | undefined;
+  port: string | undefined;
+  user: string | undefined;
+  password: string | undefined;
+  dbName: string | undefined;
+}): DataSource => {
+  for (const [key, value] of Object.entries(config)) {
```
daywiss, line 11: why do you allow for undefined values here if you are throwing if included

packages/indexer/src/database/database.provider.ts
```diff
@@ -0,0 +1,20 @@
+import { createDataSource } from "@repo/indexer-database";
+
+export async function connectToDatabase(
+  env: Record<string, string | undefined>,
+) {
+  const databaseConfig = {
```
daywiss, line 6: perhaps part of the pattern you introduce in createDataSource is to handle the uncertainty around env vars. IMO you should validate that any env vars exist before you use them, so the end user knows exactly what env var needs to be defined for the app to work, for example:
`assert(env.DATABASE_HOST,'DATABASE_HOST must be defined in env')` or something to that effect.  if you push this validation to the function params, the user has to dig through code to see what env they need to provide

daywiss (commented):
couple comments for easier maintenence

packages/indexer-api/src/database/database.provider.ts
```diff
@@ -0,0 +1,24 @@
...
+    host: env.DATABASE_HOST,
+    port: env.DATABASE_PORT,
+    user: env.DATABASE_USER,
+    password: env.DATABASE_PASSWORD,
+    dbName: env.DATABASE_NAME,
+  };
+  for (const [key, value] of Object.entries(databaseConfig)) {
+    if (!value)
+      throw new Error(`Missing required environment variable: ${key}`);
+  }
```
james-a-morris, line 16: Could we maybe use a validation library here like superstruct?

packages/indexer/src/database/database.provider.ts
melisaguevara: Updated here [e25612d](https://github.com/across-protocol/indexer/pull/8/commits/e25612db263a648fb1911b8e475acd6bfc77dfe2)

packages/indexer-api/src/main.ts
```diff
@@ -1,10 +1,29 @@
...
 
+  // Validate database config
+  const databaseConfig = {
+    host: env.DATABASE_HOST,
+    port: env.DATABASE_PORT,
+    user: env.DATABASE_USER,
+    password: env.DATABASE_PASSWORD,
+    dbName: env.DATABASE_NAME,
+  };
+  assert(databaseConfig, databaseConfigStruct);
```
daywiss, line 25: this is better, my only issue is that it doesnt tell the program runner what envs are required if something is omitted, but its ok

daywiss (approved):

packages/indexer-api/src/main.ts
daywiss: yea i understand, its the difference between saying "`missing property dbName` vs `missing property DATABASE_NAME`
its not a huge deal either way its just slightly better imo to show the env var that needs to be defined

packages/indexer-api/src/database/database.provider.ts
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    const database = await createDataSource(databaseConfig).initialize();
+    return database;
+  } catch (error) {
+    console.log("Unable to connect to database", error);
```
james-a-morris, line 8: should we add some logging here that's not the console.log?

james-a-morris (approved):
one nit

PR #9, opened 2024-08-09
feat: add deposit indexer basics

daywiss:
# motivation
we want to index deposits and other related data

## changes
this is the start of the indexer logic which adds some basics as to how this could work. This is still a work in progress:
TODO: 
- add postgres persistenence and data types
- add cached rpc provider
- general clean up

### DIFF
--- a/apps/node/src/app.ts
+++ b/apps/node/src/app.ts
@@ -6,6 +6,27 @@ import * as Indexer from "@repo/indexer";
 import * as PersistenceExample from "@repo/persistence-example";
 import * as IndexerApi from "@repo/indexer-api";
 
+import { createLogger, format, transports } from "winston";
+
+// Create the logger instance
+const logger = createLogger({
+  level: "info", // Set the default log level
+  format: format.combine(
+    format.timestamp({ format: "YYYY-MM-DD HH:mm:ss" }),
+    format.errors({ stack: true }),
+    format.printf(({ timestamp, level, message, ...meta }) => {
+      const metaString = Object.keys(meta).length
+        ? JSON.stringify(meta, null, 2)
+        : "";
+      return `[${timestamp}] ${level.toUpperCase()}: ${message} ${metaString}`;
+    }),
+  ),
+
+  transports: [
+    new transports.Console(), // Log to the console
+  ],
+});
+
 dotenv.config();
 
 async function run() {
@@ -16,8 +37,8 @@ async function run() {
       void (await Template.Main(process.env));
       return "Example template app running";
     case "indexer":
-      void (await Indexer.Main(process.env));
-      return "Indexer running";
+      void (await Indexer.Main(process.env, logger));
+      break;
     case "persistence-example":
       void (await PersistenceExample.Main(process.env));
       return "Example persistence app running";
@@ -29,4 +50,6 @@ async function run() {
   }
 }
 
-run().then(console.log).catch(console.error);
+run()
+  .then((x) => x && logger.info(x))
+  .catch(console.log);
--- a/packages/indexer/package.json
+++ b/packages/indexer/package.json
@@ -21,7 +21,15 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "superstruct": "2.0.3-1"
+    "@across-protocol/contracts": "^3.0.8",
+    "@across-protocol/sdk": "^3.1.17",
+    "@types/superstruct": "^0.8.2",
+    "ethers": "5.7.2",
+    "ioredis": "^5.4.1",
+    "redis": "^4.7.0",
+    "superstruct": "^2.0.2",
+    "typeorm": "0.3.20",
+    "winston": "^3.13.1"
   },
   "exports": {
     ".": "./src/index.ts"
--- a/packages/indexer/package.json
+++ b/packages/indexer/package.json
@@ -21,14 +21,14 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@across-protocol/constants": "^3.1.13",
     "@across-protocol/contracts": "^3.0.8",
     "@across-protocol/sdk": "^3.1.17",
     "@types/superstruct": "^0.8.2",
     "ethers": "5.7.2",
     "ioredis": "^5.4.1",
     "redis": "^4.7.0",
-    "superstruct": "^2.0.2",
-    "typeorm": "0.3.20",
+    "superstruct": "^2.0.3-1",
     "winston": "^3.13.1"
   },
   "exports": {
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -2,6 +2,7 @@ import assert from "assert";
 import * as services from "./services";
 import winston from "winston";
 import Redis from "ioredis";
+import * as acrossConstants from "@across-protocol/constants";
 
 function sleep(ms: number) {
   return new Promise((res) => setTimeout(res, ms));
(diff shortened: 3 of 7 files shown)


packages/indexer/src/services/deposits.ts
```diff
@@ -0,0 +1,70 @@
...
+    const events = await spokePool.queryFilter(
+      fundsDepositedFilter,
+      fromBlock,
+      toBlock,
+    );
+    console.log("Processing block range", fromBlock, toBlock);
+    // convert to some deposit object,
+    const deposits = events.map((event) => {
+      return {
+        inputToken: event.args.inputToken,
```
james-a-morris, line 52: In this example will we also be including transaction-adjacent information such as block number, tx hash, log index, etc

daywiss: could def include that, and probably shoudl

apps/node/src/app.ts
```diff
@@ -6,6 +6,27 @@ import * as Indexer from "@repo/indexer";
 import * as PersistenceExample from "@repo/persistence-example";
 import * as IndexerApi from "@repo/indexer-api";
 
+import { createLogger, format, transports } from "winston";
+
+// Create the logger instance
+const logger = createLogger({
```
daywiss, line 12: this needs to be updated with whatever logging settings we want, i kind of just threw stuff in here to make it useful for console logs

apps/node/src/app.ts
```diff
@@ -16,8 +37,8 @@ async function run() {
       void (await Template.Main(process.env));
       return "Example template app running";
     case "indexer":
-      void (await Indexer.Main(process.env));
-      return "Indexer running";
+      void (await Indexer.Main(process.env, logger));
+      break;
```
daywiss, line 41: passing in logger and no longer returning any end string

packages/indexer/src/main.ts
```diff
@@ -1,3 +1,109 @@
...
+    // arbitrum
+    env.INDEXER_PROVIDER_URL_42161,
+    // linea
+    env.INDEXER_PROVIDER_URL_59144,
+    // blast
+    env.INDEXER_PROVIDER_URL_81457,
+    // scroll
+    env.INDEXER_PROVIDER_URL_534352,
+    // zora
+    env.INDEXER_PROVIDER_URL_7777777,
```
daywiss, line 68: these are optional, so you can provide 1 or more to enable the clients

packages/indexer/src/main.ts
```diff
@@ -1,3 +1,109 @@
...
+  ].filter((x): x is string => !!x);
+
+  assert(
+    providerUrls.length > 0,
+    "Must provide a url for at least one provider on one chain, for example: INDEXER_PROVIDER_URL_1",
+  );
+
+  // optional redis config
+  const redisConfig =
```
daywiss, line 77: optional redis config, otherwise no persistence

packages/indexer/src/main.ts
```diff
@@ -1,3 +1,109 @@
...
+  });
+}
+
+export async function Main(
+  env: Record<string, string | undefined>,
+  logger: winston.Logger,
+) {
+  let running = true;
+  // Handle Ctrl+C (SIGINT)
+  process.on("SIGINT", async () => {
```
daywiss, line 36: added this to allow us to end the process if its looping

packages/indexer/src/main.ts
```diff
@@ -1,3 +1,109 @@
...
+    logger,
+    redis,
+  });
+
+  // do {
+  logger.info("index loop starting");
+  await depositIndexer(Date.now());
+  logger.info("index loop complete");
+  // sleep(30000);
+  // } while (running);
```
daywiss, line 105: eventually this should just loop until we signal stop somehow

packages/indexer/src/services/deposits.ts
```diff
@@ -0,0 +1,156 @@
...
+            logger,
+            redis,
+            maxBlockLookBack,
+            chainId,
+          }),
+        ];
+      }),
+    );
+
+  async function update(
```
daywiss, line 128: updates a single spoke client and saves the last block it finished on ( if redis enabled )

packages/indexer/src/services/deposits.ts
```diff
@@ -0,0 +1,156 @@
...
+    now: number,
+    chainId: number,
+    spokeClient: across.clients.SpokePoolClient,
+  ) {
+    logger.info({
+      message: "Starting update on spoke client",
+      chainId,
+    });
+    await spokeClient.update();
+    const latestBlockSearched = spokeClient.latestBlockSearched;
```
daywiss, line 138: here we can start to pull off all the state and persist it to sql once that is ready

james-a-morris:
is this ready for review?

packages/indexer/src/main.test.ts
```diff
@@ -1,16 +1,7 @@
 import { expect } from "chai";
 
-import * as main from "./main";
-
 describe("main", () => {
   it("should return true", async () => {
-    const result = await main.Main({
```
daywiss, line 7: we cant do this as a unit test, since it requires outside dependencies

packages/indexer/src/main.ts
```diff
@@ -1,23 +1,109 @@
-import { object, string, assert } from "superstruct";
-
-const databaseConfigStruct = object({
```
daywiss, line 3: didnt realize this was in the indexer, sorry will put this back in with the postgres pr

packages/indexer/package.json
```diff
@@ -21,7 +21,15 @@
...
   "dependencies": {
-    "superstruct": "2.0.3-1"
+    "@across-protocol/contracts": "^3.0.8",
+    "@across-protocol/sdk": "^3.1.17",
+    "@types/superstruct": "^0.8.2",
+    "ethers": "5.7.2",
+    "ioredis": "^5.4.1",
+    "redis": "^4.7.0",
+    "superstruct": "^2.0.2",
+    "typeorm": "0.3.20",
```
melisaguevara, line 31: I think we won't need 'typeorm' in this package as we will use the indexer-database package.

melisaguevara (commented):
lgtm! Just two nits and one comment

packages/indexer/package.json
```diff
@@ -21,7 +21,15 @@
...
   "license": "ISC",
   "dependencies": {
-    "superstruct": "2.0.3-1"
+    "@across-protocol/contracts": "^3.0.8",
+    "@across-protocol/sdk": "^3.1.17",
+    "@types/superstruct": "^0.8.2",
+    "ethers": "5.7.2",
+    "ioredis": "^5.4.1",
+    "redis": "^4.7.0",
+    "superstruct": "^2.0.2",
```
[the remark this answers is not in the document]
daywiss: my bad resolved conflict incorrectly, will fix

packages/indexer/src/services/deposits.ts
```diff
@@ -0,0 +1,156 @@
...
+    type: "postgres",
+    host: config.host,
+    port: config.port,
+    username: config.username,
+    password: config.password,
+    database: config.database,
+    logging: false,
+    entities: [],
+  });
+};
```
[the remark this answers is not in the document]
daywiss: can remove this for now

packages/indexer/src/services/deposits.ts
```diff
@@ -0,0 +1,156 @@
...
+  });
+  const spokePoolContract = new Contract(
+    address,
+    SpokePoolFactory.abi,
+    provider,
+  );
+  return new across.clients.SpokePoolClient(
+    logger,
+    spokePoolContract,
+    null,
```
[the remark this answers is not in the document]
daywiss: not avoiding, just have not gotten to it

packages/indexer/src/main.test.ts
james-a-morris: We can probably test service modules if we mock data - but I don't think this is high priority atm while we're still building the basic MVP

james-a-morris (commented):
lgtm so far

PR #11, opened 2024-08-14
feat: add config and hub clients with updating

daywiss:
# motivation
we want to add hubpool client to the spoke pool clients

# changes
This adds a new env for hubpool provider, and changes the env names for spoke pool providers. we use this to instanciate both across config and hubpool client and update them before attempting to update spoke pool clients. 

# todo
- add in postgres persistence
- retry/cache rpc provider

### DIFF
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -40,17 +40,22 @@ export async function Main(
     );
     running = false;
   });
-  const providerUrls: string[] = Object.values(
+  const spokePoolProviderUrls: string[] = Object.values(
     acrossConstants.MAINNET_CHAIN_IDs,
   )
-    .map((chainId) => env[`INDEXER_PROVIDER_URL_${chainId}`])
+    .map((chainId) => env[`INDEXER_SPOKEPOOL_PROVIDER_URL_${chainId}`])
     .filter((x): x is string => !!x);
 
   assert(
-    providerUrls.length > 0,
-    "Must provide a url for at least one provider on one chain, for example: INDEXER_PROVIDER_URL_1",
+    spokePoolProviderUrls.length > 0,
+    "Must provide a url for at least one provider on one chain, for example: INDEXER_SPOKEPOOL_PROVIDER_URL_1",
   );
 
+  assert(
+    env.INDEXER_HUBPOOL_PROVIDER_URL,
+    "requires INDEXER_HUBPOOL_PROVIDER_URL",
+  );
+  const hubPoolProviderUrl = env.INDEXER_HUBPOOL_PROVIDER_URL;
   // optional redis config
   const redisConfig =
     env.INDEXER_REDIS_HOST && env.INDEXER_REDIS_PORT
@@ -67,14 +72,17 @@ export async function Main(
   logger.info({
     message: "Starting indexer",
     redisConfig,
-    providerUrls,
+    spokePoolProviderUrls,
+    hubPoolProviderUrl,
   });
   const depositIndexer = await services.deposits.Indexer({
-    providerUrls,
+    spokePoolProviderUrls,
+    hubPoolProviderUrl,
     logger,
     redis,
   });
 
+  // TODO: add looping to keep process going
   // do {
   logger.info("index loop starting");
   await depositIndexer(Date.now());
--- a/packages/indexer/src/services/deposits.ts
+++ b/packages/indexer/src/services/deposits.ts
@@ -92,12 +92,13 @@ export async function getConfigStoreClient(params: GetConfigStoreClientParams) {
     provider,
   );
   let lastProcessedBlockNumber: number = deployedBlockNumber;
-  if (redis) {
-    lastProcessedBlockNumber = Number(
-      (await redis.get(getConfigStoreLastBlockSearchedKey(chainId))) ??
-        lastProcessedBlockNumber,
-    );
-  }
+  // for now we will always process all config store events.
+  // if (redis) {
+  //   lastProcessedBlockNumber = Number(
+  //     (await redis.get(getConfigStoreLastBlockSearchedKey(chainId))) ??
+  //       lastProcessedBlockNumber,
+  //   );
+  // }
   const eventSearchConfig = {
     fromBlock: lastProcessedBlockNumber,
     maxBlockLookBack,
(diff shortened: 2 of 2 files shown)


packages/indexer/src/main.ts
```diff
@@ -40,17 +40,22 @@ export async function Main(
     );
     running = false;
   });
-  const providerUrls: string[] = Object.values(
+  const spokePoolProviderUrls: string[] = Object.values(
```
james-a-morris, line 43: I think we can keep this as `providerUrls` because we'll be using the same RPC calls for the spoke/hub/config store

packages/indexer/src/services/deposits.ts
```diff
@@ -3,31 +3,34 @@ import {
...
-type Config = {
-  providerUrls: string[];
-  maxBlockLookBack?: number;
-  logger: winston.Logger;
-  redis: Redis | undefined;
-};
+// Maximum supported version of the configuration loaded into the Across ConfigStore.
+// It protects bots from running outdated code against newer version of the on-chain config store.
+// @dev Incorrectly setting this value may lead to incorrect behaviour and potential loss of funds.
+export const CONFIG_STORE_VERSION = 4;
```
james-a-morris, line 19: For now - we should place a link to where the main version of this is stored in the relayer

packages/indexer/src/services/deposits.ts
```diff
@@ -65,29 +68,156 @@ export async function getSpokeClient(
...
+  const address = getDeployedAddress("AcrossConfigStore", chainId);
+  const deployedBlockNumber = getDeployedBlockNumber(
+    "AcrossConfigStore",
+    chainId,
+  );
+  const configStoreContract = new Contract(
+    address,
+    AcrossConfigStoreFactory.abi,
+    provider,
+  );
```
james-a-morris, line 95: I think you can call `AcrossConfigStoreFactory.connect( )` to get a specific child class of `Contract` that has type strictness

packages/indexer/src/services/deposits.ts
```diff
@@ -65,29 +68,156 @@ export async function getSpokeClient(
...
+  let lastProcessedBlockNumber: number = deployedBlockNumber;
+  if (redis) {
+    lastProcessedBlockNumber = Number(
+      (await redis.get(getConfigStoreLastBlockSearchedKey(chainId))) ??
+        lastProcessedBlockNumber,
+    );
+  }
+  const eventSearchConfig = {
+    fromBlock: lastProcessedBlockNumber,
```
james-a-morris, line 104: What happens in the case that the config store has information contained prior to the `fromBlock`?

I.e. if the from block is 1000 and the latest inclusion in the lite chain id list is at block 999

packages/indexer/src/services/deposits.ts
```diff
@@ -123,13 +289,18 @@ export async function Indexer(config: Config) {
...
+      );
     }
   }
 
   return async function updateAll(now: number) {
     for (const [chainId, spokeClient] of spokeClientEntries) {
-      await update(now, chainId, spokeClient);
+      await updateConfigStore(now, hubPoolNetworkInfo.chainId);
+      await updateHubPool(now, hubPoolNetworkInfo.chainId);
+      await updateSpokePool(now, chainId, spokeClient);
```
james-a-morris, line 303: We don't need to update the config store + hub pool for each spoke client entry - i don't think.

Additionally - should we order:
1. update config store
2. update hub
3. in parallel, update all spokes?

james-a-morris: This file as an example of how we do this today: https://github.com/across-protocol/relayer/blob/master/src/dataworker/DataworkerClientHelper.ts

packages/indexer/src/main.ts
daywiss: The reason i split this is so you can specify the hubpool provider you want to use, if theres just a list of provider urls we cant know which one is meant for the hubpool without just assuming mainnet.  also by splitting it we could make hubpool provider optional, just like it is in the spoke client ( though for simplicity i made this required for now )

packages/indexer/src/services/deposits.ts
daywiss: this fails typescript check, its expecting generic contract `Argument of type 'AcrossConfigStore' is not assignable to parameter of type 'Contract'.`

packages/indexer/src/services/deposits.ts
daywiss: i dont actually know whats goign on under the hood so i cant say. this is probably something we need to figure out once we start deriving data from these clients

packages/indexer/src/services/deposits.ts
daywiss: > We don't need to update the config store + hub pool for each spoke client entry - i don't think.

you are right that was an error

> Additionally - should we order:
> 
> update config store
> update hub
> in parallel, update all spokes?

can make the spoke updates parallel

packages/indexer/src/services/deposits.ts
melisaguevara: @james-a-morris can we do the same for the hubpool?

packages/indexer/src/services/deposits.ts
```diff
@@ -65,29 +68,156 @@ export async function getSpokeClient(
...
+  ].join("~");
+}
+function getHubPoolLastBlockSearchedKey(chainId: number): string {
+  return [
+    "depositIndexer",
+    "hubPool",
+    "lastProcessedBlockNumber",
+    chainId,
+  ].join("~");
+}
```
melisaguevara, line 176: nit: maybe we can turn this in a single function that receives the name of the contract as a param?

melisaguevara (approved):
lgtm! Just one nit

packages/indexer/src/services/deposits.ts
daywiss: potentially!

packages/indexer/src/main.ts
james-a-morris: Ah I see - fair call in other locations we reference a constant specified in the ENV vars like `HUB_CHAIN_ID`.

packages/indexer/src/services/deposits.ts
james-a-morris: we resolved offline- we'll use the AcrossConfigStore from genesis

james-a-morris (approved):
Just the one follow up: https://github.com/across-protocol/indexer/pull/11#discussion_r1717497138

PR #12, opened 2024-08-15
feat: add basic indexer api

daywiss:
# motivation
stub out pontential api calls

# changes
hooked in deposit route to query raw deposit data. added some basic calls for mvp as well as potential future calls

# todo
add rest of data as its available in indexer in future prs

# screens
`curl localhost:8080/indexer/deposits?limit=2`
image

### DIFF
--- a/packages/indexer-api/package.json
+++ b/packages/indexer-api/package.json
@@ -27,7 +27,8 @@
     "body-parser": "^1.20.2",
     "cors": "^2.8.5",
     "express": "^4.19.2",
-    "superstruct": "2.0.3-1"
+    "superstruct": "2.0.3-1",
+    "winston": "^3.13.1"
   },
   "exports": {
     ".": "./src/index.ts"
--- a/packages/indexer-api/src/main.ts
+++ b/packages/indexer-api/src/main.ts
@@ -1,34 +1,67 @@
 import { object, string, assert } from "superstruct";
 import { ExpressApp } from "./express-app";
+import { createDataSource, DatabaseConfig } from "@repo/indexer-database";
 import * as services from "./services";
+import winston from "winston";
+import { type Router } from "express";
 
-const databaseConfigStruct = object({
-  host: string(),
-  port: string(),
-  user: string(),
-  password: string(),
-  dbName: string(),
-});
+export async function connectToDatabase(
+  databaseConfig: DatabaseConfig,
+  logger: winston.Logger,
+) {
+  try {
+    const database = await createDataSource(databaseConfig).initialize();
+    logger.info("Postgres connection established");
+    return database;
+  } catch (error) {
+    logger.error("Unable to connect to database", error);
+    throw error;
+  }
+}
+function getPostgresConfig(
+  env: Record<string, string | undefined>,
+): DatabaseConfig | undefined {
+  return env.DATABASE_HOST &&
+    env.DATABASE_PORT &&
+    env.DATABASE_USER &&
+    env.DATABASE_PASSWORD &&
+    env.DATABASE_NAME
+    ? {
+        host: env.DATABASE_HOST,
+        port: env.DATABASE_PORT,
+        user: env.DATABASE_USER,
+        password: env.DATABASE_PASSWORD,
+        dbName: env.DATABASE_NAME,
+      }
+    : undefined;
+}
 
-export async function Main(env: Record<string, string | undefined>) {
+export async function Main(
+  env: Record<string, string | undefined>,
+  logger: winston.Logger,
+) {
   const { PORT = "8080" } = env;
   const port = Number(PORT);
 
-  // Validate database config
-  const databaseConfig = {
-    host: env.DATABASE_HOST,
-    port: env.DATABASE_PORT,
-    user: env.DATABASE_USER,
-    password: env.DATABASE_PASSWORD,
-    dbName: env.DATABASE_NAME,
-  };
-  assert(databaseConfig, databaseConfigStruct);
+  const postgresConfig = getPostgresConfig(env);
+  const postgres = postgresConfig
+    ? await connectToDatabase(postgresConfig, logger)
+    : undefined;
 
   const exampleRouter = services.example.getRouter();
-  const app = ExpressApp({ example: exampleRouter });
+  const allServices: Record<string, Router> = {
+    example: exampleRouter,
+  };
+  if (postgres) {
+    const indexerRouter = services.indexer.getRouter(postgres);
+    allServices.indexer = indexerRouter;
+  }
+  const app = ExpressApp(allServices);
 
+  logger.info({
+    message: `Starting indexer api on port ${port}`,
+  });
   void (await new Promise((res) => {
     app.listen(port, () => res(app));
   }));
-  console.log(`Indexer api listening on port ${port}`);
 }
--- a/packages/indexer-api/src/services/indexer.ts
+++ b/packages/indexer-api/src/services/indexer.ts
@@ -1,0 +1,179 @@
+import assert from "assert";
+import * as s from "superstruct";
+import { Request, Response, NextFunction, Router } from "express";
+import { JSON } from "../types";
+import { DataSource, entities } from "@repo/indexer-database";
+
+type APIHandler = (
+  params?: JSON,
+) => Promise<JSON> | JSON | never | Promise<never> | void | Promise<void>;
+export function Indexer(db: DataSource): Record<string, APIHandler> {
+  const DepositParams = s.object({
+    id: s.string(),
+  });
+  // get a single deposit by an id, regardless of chain
+  async function deposit(params: JSON) {
+    s.assert(params, DepositParams);
+    throw new Error(`Deposit id not found ${params.id}`);
+  }
+  // possible filter options
+  const DepositsParams = s.object({
+    depositor: s.optional(s.string()),
+    recipient: s.optional(s.string()),
+    inputToken: s.optional(s.string()),
+    outputToken: s.optional(s.string()),
+    integrator: s.optional(s.string()),
+    status: s.optional(s.string()),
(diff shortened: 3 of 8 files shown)


packages/indexer-api/package.json
```diff
@@ -27,7 +27,8 @@
     "body-parser": "^1.20.2",
     "cors": "^2.8.5",
     "express": "^4.19.2",
-    "superstruct": "2.0.3-1"
+    "superstruct": "2.0.3-1",
```
james-a-morris, line 30: ooc what is `2.0.3-1`? Is this a beta build

packages/indexer-api/src/main.ts
```diff
@@ -1,34 +1,67 @@
 import { object, string, assert } from "superstruct";
 import { ExpressApp } from "./express-app";
+import { createDataSource, DatabaseConfig } from "@repo/indexer-database";
 import * as services from "./services";
+import winston from "winston";
+import { type Router } from "express";
```
james-a-morris, line 6: I'm not sure I've seen this syntax before - is this importing the router as a `type`?

packages/indexer-api/src/main.ts
```diff
@@ -1,34 +1,67 @@
...
+        user: env.DATABASE_USER,
+        password: env.DATABASE_PASSWORD,
+        dbName: env.DATABASE_NAME,
+      }
+    : undefined;
+}
 
-export async function Main(env: Record<string, string | undefined>) {
+export async function Main(
+  env: Record<string, string | undefined>,
```
james-a-morris, line 40: If we're including superstruct does it make sense to do this parsing prior to entering the `Main` function and exporting a superstruct type?

james-a-morris: This a more long term item but it could be useful to start reasoning about it now

packages/indexer-api/src/main.ts
```diff
@@ -1,34 +1,67 @@
...
+  const postgres = postgresConfig
+    ? await connectToDatabase(postgresConfig, logger)
+    : undefined;
 
   const exampleRouter = services.example.getRouter();
-  const app = ExpressApp({ example: exampleRouter });
+  const allServices: Record<string, Router> = {
+    example: exampleRouter,
+  };
+  if (postgres) {
```
james-a-morris, line 55: If postgres fails - would we want to throw?

packages/indexer-api/src/services/indexer.ts
```diff
@@ -0,0 +1,179 @@
+import assert from "assert";
+import * as s from "superstruct";
+import { Request, Response, NextFunction, Router } from "express";
+import { JSON } from "../types";
+import { DataSource, entities } from "@repo/indexer-database";
+
+type APIHandler = (
```
james-a-morris, line 7: Is this generic? Can we place it in a common types directory?

packages/indexer-api/src/services/indexer.ts
```diff
@@ -0,0 +1,179 @@
...
+        const method = api[action];
+        //check if it exists
+        if (method) {
+          // call and return result
+          const result = await method(params);
+          return res.json(result);
+        }
+        throw new Error(`Unknown api call: ${action}`);
+      } catch (err) {
+        next(err);
```
james-a-morris, line 173: OOC what is the purpose of calling `next` directly? Is this so that we capture the error to a specific error handler?

packages/indexer-api/package.json
daywiss: dont know, i think its just matching the version we had used elsewhere

packages/indexer-api/src/main.ts
daywiss: yes its a way you can specify you are only importing the type per variable

packages/indexer-api/package.json
melisaguevara: Yes, I mentioned it in a previous PR.
It is [this pre-release version](https://github.com/ianstormtaylor/superstruct/releases/tag/v2.0.3-1) that fixes superstruct exports. I was having importing issues with latest.

packages/indexer-api/src/main.ts
daywiss: yes its possible, id probably suggest exporting a single envToConfig function which takes an env and produces and validates the config. and that gets passed into the library from the node/app

packages/indexer-api/src/main.ts
daywiss: it will throw on `connectToDatabase` a few lines up.  this is just so postgres can be optional, in case you are not running an api that needs it

packages/indexer-api/src/services/indexer.ts
daywiss: not sure it should be enshrined yet, was just trying it as a general interface to api calls, there are some drawbacks to it since you have to cast your return types to json, and you have to validate your input parameters, but might not be a way around this

packages/indexer-api/src/services/indexer.ts
```diff
@@ -0,0 +1,179 @@
...
+    skip: s.optional(
+      s.coerce(s.number(), s.string(), (value) => parseInt(value)),
+    ),
+    // pagination limit, how many to return after the start, note we convert string to number
+    limit: s.optional(
+      s.coerce(s.number(), s.string(), (value) => parseInt(value)),
+    ),
+  });
+  // get a list of deposits filtered by options
+  async function deposits(queryParams: JSON) {
```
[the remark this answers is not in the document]
daywiss: can you suggest a quick way to do that?

packages/indexer-api/src/services/indexer.ts
daywiss: you dont want to throw in an express handler, that will crash the app, passing this to next lets a successive handler deal with it, and in this case if you look at the express file, theres an error handler at the end which will parse and return it to the caller

packages/indexer-api/src/services/indexer.ts
```diff
@@ -0,0 +1,179 @@
...
+      queryBuilder.andWhere("deposit.outputToken = :outputToken", {
+        outputToken: params.outputToken,
+      });
+    }
+
+    if (params.integrator) {
+      queryBuilder.andWhere("deposit.integrator = :integrator", {
+        integrator: params.integrator,
+      });
+    }
```
melisaguevara, line 71: We don't have this in the Deposit entity but I imagine we would like to have it.
@james-a-morris do you know where should we get the integrator from?

melisaguevara (approved):
LGTM!

packages/indexer-api/src/services/indexer.ts
```diff
@@ -0,0 +1,179 @@
+import assert from "assert";
+import * as s from "superstruct";
+import { Request, Response, NextFunction, Router } from "express";
+import { JSON } from "../types";
+import { DataSource, entities } from "@repo/indexer-database";
+
+type APIHandler = (
+  params?: JSON,
+) => Promise<JSON> | JSON | never | Promise<never> | void | Promise<void>;
+export function Indexer(db: DataSource): Record<string, APIHandler> {
```
amateima, line 10: It is not very easy to tell what this function does only by looking at its name. Also, putting all the endpoints handlers in a single function creates a huge function that is not very easy to read.

packages/indexer-api/src/services/indexer.ts
```diff
@@ -0,0 +1,179 @@
...
+  const DepositParams = s.object({
+    id: s.string(),
+  });
```
amateima, line 13: Maybe this type of Data Transfer Object (url params, query parameters, request body, etc) should have their own place and since this is something related to the REST API, my feeling is that this should be part of the `entrypoint` component I mentioned above

packages/indexer-api/src/services/indexer.ts
```diff
@@ -0,0 +1,179 @@
...
+  // get a single deposit by an id, regardless of chain
+  async function deposit(params: JSON) {
+    s.assert(params, DepositParams);
+    throw new Error(`Deposit id not found ${params.id}`);
+  }
```
amateima, line 18: I think this is the only piece of code (the endpoint handler) that should stay here in the service file. Declaring endpoints, query parameters, url parameters, etc should be put in a single place from where the service will be called

packages/indexer-api/src/services/indexer.ts
amateima: We need the transaction call data to extract the integrator. Not sure if there is smth useful in the SDK that we can use

packages/indexer-api/src/services/indexer.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// build up express style calls to our example api
+export function getRouter(db: DataSource): Router {
+  const router = Router();
+  const api = Indexer(db);
+  // example call: curl localhost:8080/example/now -> timestamp
+  router.get(
+    "/:action",
+    async (req: Request, res: Response, next: NextFunction) => {
```
amateima, line 157: Since this API will be exposed and promoted publicly, I think we should go for a more "RESTful" and declarative way of specifying endpoints, instead of this procedure call style.
Also it's easier to read the code if we mention explicitly the available endpoints: "deposit", "deposits", etc

amateima (changes requested):

james-a-morris (commented):
can we include some form of swagger endpoint? AFAICT it should be decorators

amateima:
> can we include some form of swagger endpoint? AFAICT it should be decorators

I think swagger documentation is out of the scope of this PR, at least for now. It's possible that some of these endpoints will be documented on the docs.across.to website, so in this case a potential swagger UI will be used mostly internally

daywiss:
@amateima @james-a-morris Have several follow on prs to this which will address many of these comments, would like to get this in to no block new work for now

amateima (approved):

PR #13, opened 2024-08-15
feat: including sql persistence for deposits, fills and slow fill requests

melisaguevara:
# motivation
we want to persist spoke pool events on postgresql

# changes
- packages/indexer-database:
    - Including entity definition and its related migration for: Deposit, Fill, SlowFillRequest
- packages/indexer:
    - Including functions to format and persist Deposits, Fills and SlowFillRequests

# todo
- handle errors when insertions violate unique constraints
- get migrations to run automatically when running indexer app. In the meantime, I'm including instructions on how to do this in packages/indexer/readme.md.
- rename deposit indexer to spoke indexer?

### DIFF
--- a/packages/indexer-database/migrations.config.ts
+++ b/packages/indexer-database/migrations.config.ts
@@ -1,4 +1,7 @@
+import dotenv from "dotenv"
 import { DataSource } from "typeorm"
+
+dotenv.config();
 
 export const AppDataSource = new DataSource({
     type: "postgres",
--- a/packages/indexer-database/migrations.config.ts
+++ b/packages/indexer-database/migrations.config.ts
@@ -1,18 +1,17 @@
-import dotenv from "dotenv"
-import { DataSource } from "typeorm"
+import dotenv from "dotenv";
+import { DataSource } from "typeorm";
 
 dotenv.config();
 
 export const AppDataSource = new DataSource({
-    type: "postgres",
-    host: process.env.DATABASE_HOST,
-    port: parseInt(process.env.DATABASE_PORT || "5432", 10),
-    username: process.env.DATABASE_USER,
-    password: process.env.DATABASE_PASSWORD,
-    database: process.env.DATABASE_NAME,
-    logging: "all",
-    entities: ["src/entities/*.ts"],
-    migrationsTableName: "_migrations",
-    migrations: ["src/migrations/*.ts"],
+  type: "postgres",
+  host: process.env.DATABASE_HOST,
+  port: parseInt(process.env.DATABASE_PORT || "5432", 10),
+  username: process.env.DATABASE_USER,
+  password: process.env.DATABASE_PASSWORD,
+  database: process.env.DATABASE_NAME,
+  logging: "all",
+  entities: ["src/entities/*.ts"],
+  migrationsTableName: "_migrations",
+  migrations: ["src/migrations/*.ts"],
 });
-
--- a/packages/indexer-database/package.json
+++ b/packages/indexer-database/package.json
@@ -24,6 +24,8 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@across-protocol/sdk": "^3.1.17",
+    "dotenv": "^16.4.5",
     "pg": "^8.4.0",
     "reflect-metadata": "^0.1.13",
     "superstruct": "2.0.3-1",
--- a/packages/indexer-database/package.json
+++ b/packages/indexer-database/package.json
@@ -24,8 +24,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@across-protocol/sdk": "^3.1.17",
-    "dotenv": "^16.4.5",
+    "@across-protocol/sdk": "^3.1.25",
     "pg": "^8.4.0",
     "reflect-metadata": "^0.1.13",
     "superstruct": "2.0.3-1",
@@ -43,6 +42,7 @@
     "@types/node": "^16.11.10",
     "@types/pg": "^8.11.6",
     "chai": "^4.5.0",
+    "dotenv": "^16.4.5",
     "eslint": "^8.57.0",
     "mocha": "^10.7.0",
     "nyc": "^17.0.0",
--- a/packages/indexer-database/src/entities/deposit.entity.ts
+++ b/packages/indexer-database/src/entities/deposit.entity.ts
@@ -1,0 +1,99 @@
+import {
+  Column,
+  CreateDateColumn,
+  Entity,
+  PrimaryGeneratedColumn,
+  Unique,
+  UpdateDateColumn,
+} from "typeorm";
+
+export type DepositStatus =
+  | "unfilled"
+  | "filled"
+  | "slowFillRequested"
+  | "slowFilled"
+  | "expired"
+  | "refunded";
+
+// TODO: Add expiredRefundBundle and slowFillBundle when we have the Bundle entity
+@Entity()
+@Unique("UK_deposit_depositId_originChainId", ["depositId", "originChainId"])
+export class Deposit {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column()
+  uuid: string;
+
+  @Column()
+  depositId: number;
+
+  @Column()
+  originChainId: number;
+
+  @Column()
+  destinationChainId: number;
+
+  @Column()
+  fromLiteChain: boolean;
+
+  @Column()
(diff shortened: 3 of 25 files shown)


packages/indexer-database/src/entities/deposit.entity.ts
```diff
@@ -0,0 +1,99 @@
...
+  outputAmount: string;
+
+  @Column()
+  message: string;
+
+  @Column()
+  exclusiveRelayer: string;
+
+  @Column({ nullable: true })
+  exclusivityDeadline?: Date;
```
melisaguevara, line 68: @james-a-morris you asked me before if nullable was defaulted to false. It is.

packages/indexer-database/src/entities/deposit.entity.ts
```diff
@@ -0,0 +1,99 @@
...
+  | "slowFilled"
+  | "expired"
+  | "refunded";
+
+// TODO: Add expiredRefundBundle and slowFillBundle when we have the Bundle entity
+@Entity()
+@Unique("UK_deposit_depositId_originChainId", ["depositId", "originChainId"])
+export class Deposit {
+  @PrimaryGeneratedColumn()
+  id: number;
```
daywiss, line 23: does this mean sql will generate an incrementing id? im assuming maybe theres other ways to prevent dupe entries?

packages/indexer-database/src/migrations/1723753792975-Deposit.ts
```diff
@@ -0,0 +1,44 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Deposit1723753792975 implements MigrationInterface {
+  name = "Deposit1723753792975";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
```
daywiss, line 6: what is this?

packages/indexer/src/main.ts
```diff
@@ -69,9 +89,16 @@ export async function Main(
     ? await initializeRedis(redisConfig, logger)
     : undefined;
 
+  // optional postgresConfig
+  const postgresConfig = getPostgresConfig(env);
+  const postgres = postgresConfig
```
daywiss, line 94: fyi if this is defined you should add a cleanup routine to close connections ( if needed ) at end of main near line 121

daywiss (commented):
this is looking really good

packages/indexer-database/src/migrations/1723753792975-Deposit.ts
melisaguevara: This is autogenerated by TypeORM. Is what it uses to keep track of executed and unexecuted migrations.

packages/indexer-database/src/entities/deposit.entity.ts
melisaguevara: > does this mean sql will generate an incrementing id?
Yes

> im assuming maybe theres other ways to prevent dupe entries?
The "depositId" field used in the unique clause is the property of the deposit event though.
I think other options could be using only the "uuid" or a composite key of "uuid" + "transactionHash" for example.

packages/indexer-database/migrations.config.ts
```diff
@@ -1,5 +1,8 @@
+import dotenv from "dotenv"
 import { DataSource } from "typeorm"
 
+dotenv.config();
+
```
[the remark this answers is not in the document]
melisaguevara: I get your point. But I don't think this is meant to be an app. The reason i'm including this is that we should be able to generate and run migrations using the pnpm scripts. Those commands need the connection to the database, so I needed direct access to the env vars. But let me see if I can think of an alternative solution.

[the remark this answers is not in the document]
james-a-morris: Will this be run outside of a CI with the ENV already set?

packages/indexer-database/package.json
```diff
@@ -24,6 +24,8 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@across-protocol/sdk": "^3.1.17",
```
james-a-morris, line 27: I think we should bump this to latest

packages/indexer-database/src/entities/deposit.entity.ts
james-a-morris: Is this purely an internal tracking mechanism? I thought we spoke about how the PK would be some unique combination of existing fields?

packages/indexer-database/src/entities/deposit.entity.ts
```diff
@@ -0,0 +1,99 @@
...
+  | "unfilled"
+  | "filled"
+  | "slowFillRequested"
+  | "slowFilled"
+  | "expired"
+  | "refunded";
+
+// TODO: Add expiredRefundBundle and slowFillBundle when we have the Bundle entity
+@Entity()
+@Unique("UK_deposit_depositId_originChainId", ["depositId", "originChainId"])
```
james-a-morris, line 20: This constraint will fail for reorgs

packages/indexer/README.md
```diff
@@ -2,7 +2,10 @@
 This package is meant to read data from an rpc provider, modify it as necessary and insert into a database.
 
 ## Run
-To run, go to apps/node and run `APP=indexer pnpm start`
+To run:
+
+1. Go to packages/indexer-database and run `pnpm db:migration:run`. When running without docker, set the database config on packages/indexer-database/.env
```
james-a-morris, line 7: Is there a way to do this all from the TLD?

james-a-morris (commented):
few comments

packages/indexer-database/migrations.config.ts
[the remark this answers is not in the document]
melisaguevara: > Will this be run outside of a CI with the ENV already set?

Working locally without docker, for example

packages/indexer/README.md
melisaguevara: Yes, I will include the scripts

packages/indexer-database/src/entities/deposit.entity.ts
melisaguevara: > Is this purely an internal tracking mechanism?

yes

>  I thought we spoke about how the PK would be some unique combination of existing fields?

in the tech specs we had an "id" for each entity so I kept that. But should we use composite primary keys instead of unique constraints?

packages/indexer-database/src/entities/deposit.entity.ts
melisaguevara: I'm not sure how to handle this.
- Should we change the unique constraint for "uuid"?
- In case of a reorg, should we keep the invalidated deposits/fills/etc and flag them somehow or should we delete them?

daywiss (approved):

packages/indexer/src/main.ts
melisaguevara: Added here [285678c](https://github.com/across-protocol/indexer/pull/13/commits/285678c5c3fedadf76583501b232b1f6f71809f9)

packages/indexer-database/package.json
melisaguevara: Updated here [c7ef422](https://github.com/across-protocol/indexer/pull/13/commits/c7ef422dfaaf8c8664ad2a1cc0bf17a6fc3f7bf2)

packages/indexer-database/src/entities/deposit.entity.ts
james-a-morris: That's fair - I was under the assumption the uuid was a composite key (potentially strung together)

packages/indexer-database/src/entities/deposit.entity.ts
james-a-morris: Do we need a uniqueness constraint given that we have a PK already?

packages/indexer/README.md
melisaguevara: They were already there, my bad.
But I included the instructions to run them here [ee42f59](https://github.com/across-protocol/indexer/pull/13/commits/ee42f59ca82ade33d81e8c78d34c6291cd4f9956)

packages/indexer-database/src/entities/deposit.entity.ts
melisaguevara: As discussed offline, we will keep both the id (as PK) and the unique constraint.

james-a-morris (approved):
Thanks for reviewing my comments

PR #14, opened 2024-08-21
fix: modify entities to save evm events

melisaguevara:
This PR:
- Deletes the entities Deposit, Fill and SlowFillRequest that were created before.
- Creates a new schema 'evm' for storing events from evm compatible chains.
- Creates new entities representing some of the Across V3 contract events: V3FundsDeposited, FilledV3Relay and RequestedV3SlowFill.


packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
```diff
@@ -7,18 +7,12 @@ import {
...
-@Unique("UK_deposit_depositId_originChainId", ["depositId", "originChainId"])
-export class Deposit {
+@Entity({ schema: "evm" })
+@Unique("UK_v3FundsDeposited_depositId_originChainId", [
+  "depositId",
+  "originChainId",
+])
+export class V3FundsDeposited {
   @PrimaryGeneratedColumn()
   id: number;
```
amateima, line 18: Can we be more specific about the uuid column and name it smth like relayHash?

packages/indexer-database/src/migrations/1724271636924-V3FundsDeposited.ts
```diff
@@ -0,0 +1,45 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
```
amateima, line 1: I think we can remove the previous migration files and any references to the old table names as the app was not deployed yet and we don't have any database to maintain

packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
melisaguevara: Renamed here [a2a2667](https://github.com/across-protocol/indexer/pull/14/commits/a2a2667cb4394afd98f602d53f21e30e1d284a16)

packages/indexer-database/src/migrations/1724271636924-V3FundsDeposited.ts
melisaguevara: Deleted here [a2a2667](https://github.com/across-protocol/indexer/pull/14/commits/a2a2667cb4394afd98f602d53f21e30e1d284a16)

daywiss (approved):
lgtm

PR #16, opened 2024-08-23
feat: add optional retry provider to indexer

daywiss:


packages/indexer/src/services/deposits.ts
```diff
@@ -185,10 +214,26 @@ export async function Indexer(config: Config) {
...
-  const hubPoolProvider = new providers.JsonRpcProvider(hubPoolProviderUrl);
-  const hubPoolNetworkInfo = await hubPoolProvider.getNetwork();
+  let redisCache = undefined;
+  if (redis && retryProvider) {
+    redisCache = new RedisCache(redis);
+  }
+  // This is weird but we need to get the chain id from the provider, before calling the retry provider
+  const tempProvider = new providers.JsonRpcProvider(hubPoolProviderUrl);
+  const hubPoolNetworkInfo = await tempProvider.getNetwork();
+  const hubPoolProvider =
```
daywiss, line 227: made the retryprovider optional, just because its a pain to have to specify all the envs if you are just developing

packages/indexer/package.json
```diff
@@ -23,8 +23,8 @@
   "dependencies": {
     "@across-protocol/constants": "^3.1.13",
     "@across-protocol/contracts": "^3.0.8",
-    "@across-protocol/sdk": "^3.1.25",
-    "ethers": "5.7.2",
+    "@across-protocol/sdk": "^3.1.27",
+    "ethers": "^5.7.2",
```
james-a-morris, line 27: OOC why are we pegging this version?

packages/indexer/src/main.ts
```diff
@@ -5,6 +5,7 @@ import Redis from "ioredis";
 import * as acrossConstants from "@across-protocol/constants";
 import { DatabaseConfig } from "@repo/indexer-database";
 import { connectToDatabase } from "./database/database.provider";
+import * as s from "superstruct";
```
james-a-morris, line 8: We have this logic below `sleep` as `delay` in the SDK btw

packages/indexer/src/main.ts
```diff
@@ -48,18 +49,52 @@ function getPostgresConfig(
     : undefined;
 }
 
+// superstruct coersion to turn string into an int and validate
+const stringToInt = s.coerce(s.number(), s.string(), (value) =>
+  parseInt(value),
+);
+function getRetryProviderConfig(
```
james-a-morris, line 56: Could be helpful: https://github.com/across-protocol/relayer/blob/master/src/utils/ProviderUtils.ts

packages/indexer/src/main.ts
```diff
@@ -48,18 +49,52 @@ function getPostgresConfig(
...
-  });
+  //   let running = true;
+  //   // Handle Ctrl+C (SIGINT)
+  //   dont need this until we have some kind of looping
+  //   process.on("SIGINT", async () => {
+  //     logger.info(
+  //       "SIGINT received. Please wait or CTRL + c again to forcefully exit...",
+  //     );
+  //     running = false;
+  //   });
```
james-a-morris, line 97: OOC is this meant to be in the code?

packages/indexer/src/redisCache.ts
```diff
@@ -0,0 +1,16 @@
...
+  async get<ObjectType>(key: string): Promise<ObjectType | null> {
+    return this.redis.get(key) as ObjectType | null;
+  }
+  async set<ObjectType>(
+    key: string,
+    value: ObjectType,
+  ): Promise<string | undefined> {
+    await this.redis.set(key, JSON.stringify(value));
+    return undefined;
+  }
```
james-a-morris, line 15: Is this meant to return undefined always? Could we do the following:
```suggestion
  set<ObjectType>(
    key: string,
    value: ObjectType,
  ): Promise<string | undefined> {
    return this.redis.set(key, JSON.stringify(value));
  }
```

packages/indexer/src/services/deposits.ts
```diff
@@ -97,13 +98,6 @@ export async function getConfigStoreClient(params: GetConfigStoreClientParams) {
     provider,
   );
   let lastProcessedBlockNumber: number = deployedBlockNumber;
-  // for now we will always process all config store events.
-  // if (redis) {
-  //   lastProcessedBlockNumber = Number(
-  //     (await redis.get(getLastBlockSearchedKey('configStore',chainId))) ??
-  //       lastProcessedBlockNumber,
-  //   );
-  // }
```
james-a-morris, line 106: Was this holdover code?

packages/indexer/src/services/deposits.ts
```diff
@@ -167,6 +161,40 @@ function getLastBlockSearchedKey(
...
+};
+type RetryProviderDeps = {
+  cache: across.interfaces.CachingMechanismInterface;
+  logger: winston.Logger;
+  providerUrl: string;
+  chainId: number;
+};
+function getRetryProvider(params: RetryProviderConfig & RetryProviderDeps) {
+  return new across.providers.RetryProvider(
+    [[params.providerUrl], [undefined]],
```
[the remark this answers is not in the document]
james-a-morris: I'd check out https://github.com/across-protocol/relayer/blob/master/src/utils/ProviderUtils.ts for more info on a concrete example

pnpm-lock.yaml
james-a-morris, line 1: Nit: this is a big diff - any chance we can minimize it given that the package.json changes are relative small?

james-a-morris (commented):
a few comments. great start!

packages/indexer/package.json
daywiss: matching the version in the across sdk exactly

packages/indexer/src/main.ts
daywiss: sorry which logic?

packages/indexer/src/main.ts
daywiss: i can remove it, it was there to be able to cleanly exit the app if we were in a loop, but theres no loop yet so i removed it

pnpm-lock.yaml
daywiss: looks like this got fixed after open pr was merged

packages/indexer/src/main.ts
daywiss: oh i see the functions under this

packages/indexer/src/services/deposits.ts
daywiss: yes since we dont need to persist last block for config store. will remove

packages/indexer/src/services/deposits.ts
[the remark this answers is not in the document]
daywiss: this shows the constructor taking `params: ConstructorParameters<typeof ethers.providers.StaticJsonRpcProvider>[]`
which has like an extra `[]` at the end, which i thought was a bit strange since `ConstructorParameters` already returns an array of arguments

packages/indexer/src/main.ts
daywiss: removed it for now

packages/indexer/src/services/deposits.ts
```diff
@@ -175,6 +203,7 @@ type Config = {
   logger: winston.Logger;
   redis: Redis | undefined;
   postgres: DataSource | undefined;
+  retryProvider?: RetryProviderConfig;
```
amateima, line 206:
```suggestion
  retryProviderConfig?: RetryProviderConfig;
```

amateima (approved):

packages/indexer/src/main.ts
```diff
@@ -48,18 +49,52 @@ function getPostgresConfig(
...
+          stringToInt,
+        ),
+        noTtlBlockDistance: s.create(env.NO_TTL_BLOCK_DISTANCE, stringToInt),
+        providerCacheTtl: s.create(env.PROVIDER_CACHE_TTL, stringToInt),
+        nodeQuorumThreshold: s.create(env.NODE_QUORUM_THRESHOLD, stringToInt),
+        retries: s.create(env.RETRIES, stringToInt),
+        delay: s.create(env.DELAY, stringToInt),
+      }
+    : undefined;
+}
```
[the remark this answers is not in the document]
melisaguevara: ^ I agree, it would be easier to handle.

melisaguevara (approved):
lgtm!

packages/indexer/src/main.ts
```diff
@@ -48,18 +45,43 @@ function getPostgresConfig(
...
+): services.deposits.RetryProviderConfig | undefined {
+  return env.PROVIDER_CACHE_NAMESPACE &&
+    env.MAX_CONCURRENCY &&
+    env.PCT_RPC_CALLS_LOGGED &&
+    env.STANDARD_TTL_BLOCK_DISTANCE &&
+    env.NO_TTL_BLOCK_DISTANCE &&
+    env.PROVIDER_CACHE_TTL &&
+    env.NODE_QUORUM_THRESHOLD &&
+    env.RETRIES &&
+    env.DELAY
```
james-a-morris, line 63: I think this may be easier to implement in superstruct?

packages/indexer/src/redisCache.ts
```diff
@@ -0,0 +1,15 @@
+import * as across from "@across-protocol/sdk";
+import Redis from "ioredis";
+
+export class RedisCache implements across.interfaces.CachingMechanismInterface {
+  constructor(private redis: Redis) {}
+  async get<ObjectType>(key: string): Promise<ObjectType | null> {
+    return this.redis.get(key) as ObjectType | null;
```
james-a-morris, line 7: does this need to be typecast? I.e. can we supply the type in the `get` generic?

packages/indexer/src/main.ts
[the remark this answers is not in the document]
daywiss: yea wasnt sure how to do this, but let me add a task to refactor this https://github.com/across-protocol/indexer/pull/16#discussion_r1731208661

packages/indexer/src/main.ts
daywiss: ive got a task to refactor this in new pr david/acx-2599-refactor-retry-provider-configs-to-use-constants-rather-than

packages/indexer/src/redisCache.ts
daywiss: not sure, i get this: `Type 'string | null' is not assignable to type 'ObjectType | null'.`

PR #17, opened 2024-08-26
improve(docker): dockerfiles

james-a-morris:
This reduces the need for multiple docker commands in the package.json and allows for just calling `docker-compose up`

Requirements:

ENV:  
    - for redis the host URL must be `redis` in your ENV when using `docker-compose`
    - for postgres the host URL must be `postgres` in your ENV when using `docker-compose`

To build:
    - `docker build -t across-indexer-image . `
 To Run:
     - `docker run -eAPP=indexer --env-file ./.env -it across-indexer-image`
     - `docker run -eAPP=indexeri-api --env-file ./.env -it across-indexer-image`

Additional comments:

If selecting `indexer-api` then the production docker runtime will attempt to run migrations.


docker-compose.yml
```diff
@@ -4,40 +4,62 @@ services:
...
-      POSTGRES_PASSWORD: password
-      POSTGRES_DB: mydatabase
+      TZ: GMT
+      POSTGRES_USER: ${DATABASE_USER}
+      POSTGRES_PASSWORD: ${DATABASE_PASSWORD}
+      POSTGRES_DB: ${DATABASE_NAME}
+      PG_DATA: /var/lib/postgresql/data
     volumes:
-      - ./volumes/postgres:/var/lib/postgresql/data
+      - indexer-pgdata:/var/lib/postgresql/data
```
amateima, line 13: I like named volumes instead of volumes stored on disk 👍

docker/dev.command.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+pnpm db:indexer-database:migrate:run
+pnpm --filter @repo/node-app start
```
amateima, line 3: Which app will be started with this command?

docker-compose.yml
```diff
@@ -4,40 +4,62 @@ services:
...
+    build:
+      context: .
+      target: development
+    container_name: indexer_api
+    volumes:
+      - .:/usr/src/app
+      - indexer-node-modules:/usr/src/app/node_modules
+    ports:
+      - ${PORT}:${PORT}
+    command: ["/bin/sh", "./docker/dev.command.sh"]
```
amateima, line 53: Both the indexer and the indexer api will run this script resulting in the migration files to be executed twice, so we need to differentiate the commands executed by these 2 services. Different sh files or commands specified directly in the `command` instruction

amateima (changes requested):
Screenshot 2024-08-27 at 02 37 53

I tried to build the image locally by running `docker build . -t across-indexer-image` and got the error from the screenshot.

Also, I think we need to do some things that are particular for monorepos. Didn't read much of it, but I found [this](https://turbo.build/repo/docs/guides/tools/docker) in the Turborepo docs 👀

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM node:lts-iron AS development
```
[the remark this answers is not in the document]
james-a-morris: Updated. `lts/iron` is `v20` on the latest minor version btw

docker-compose.yml
james-a-morris: https://github.com/across-protocol/indexer/pull/17/commits/a3f7c462260bcb515c7f8b74413d8ebc602f0c1b

docker-compose.yml
james-a-morris: These are differentiated on 
```
environment:
      APP: indexer
```
and 
```
environment:
    APP: indexer-api
```

docker-compose.yml
james-a-morris: disregard above.

daywiss (approved):

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+FROM node:lts-iron AS production
+
+ARG NODE_ENV=production
+ENV NODE_ENV=${NODE_ENV}
+
+WORKDIR /usr/src/app
+COPY package*.json ./
+COPY pnpm-lock.yaml ./
+RUN pnpm install --frozen-lockfile --ignore-scripts
+COPY . .
```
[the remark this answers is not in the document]
james-a-morris: We no longer need to do this :)

docker/dev.command.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Check if the APP ENV is indexer-api
+if [ "$APP" = "indexer-api" ]; then
```
amateima, line 4: Here should be the indexer app

docker/prod.command.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# This should be loaded in the apps/node directory
+
+# Check if the APP ENV is indexer-api
+if [ "$APP" = "indexer-api" ]; then
```
amateima, line 6:
```suggestion
if [ "$APP" = "indexer" ]; then
```

docker/prod.command.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# This should be loaded in the apps/node directory
+
+# Check if the APP ENV is indexer-api
```
amateima, line 5:
```suggestion
# Check if the APP ENV is indexer
```

docker/prod.command.sh
```diff
@@ -0,0 +1,12 @@
...
+if [ "$APP" = "indexer-api" ]; then
+    cd ../../packages/indexer-database
+    npm run typeorm migration:run -- --dataSource dist/migrations.config.js
+    cd ../../apps/node
+fi
+
+node dist/app.js
```
amateima, line 12: Do you think would be a more readable code if we use pnpm commands here?

packages/indexer-api/package.json
```diff
@@ -31,7 +31,7 @@
     "winston": "^3.13.1"
   },
   "exports": {
-    ".": "./src/index.ts"
+    ".": "./dist/index.js"
```
amateima, line 34: I guess this is related to how we use and import packages, but what's more exactly the impact of this change?

melisaguevara (commented):
Should we add the ENV requirements and the run/build commands to the README?

docker/dev.command.sh
melisaguevara: @amateima do you see any issues if we delete this condition?
I think it's ok that we try to run migrations when running any of the apps. If migrations were already executed the command will do nothing.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+COPY --from=development /usr/src/app/apps ./apps
+COPY --from=development /usr/src/app/packages ./packages
+COPY --from=development /usr/src/app/node_modules ./node_modules
+
+# Change the working directory to the apps/node directory
+WORKDIR /usr/src/app/apps/node
+
+COPY ./docker ./docker
+
+CMD ["sh", "./docker/prod.command.sh"]
```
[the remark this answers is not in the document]
james-a-morris: https://github.com/across-protocol/indexer/pull/17/commits/57b5c9989d94a5d88b370feb4a59321b294e7ebf

.dockerignore
```diff
@@ -0,0 +1,7 @@
+node_modules
+Dockerfile
+.dockerignore
+.vscode
+.pnpm-store
+.turbo
+**/dist
```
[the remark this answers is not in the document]
james-a-morris: We could - but the Dockerfile should only be migrating source code.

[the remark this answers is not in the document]
james-a-morris: makes sense +1 - https://github.com/across-protocol/indexer/pull/17/commits/3eeac46df68e98572b2bae5c3c6a86c1c3e198e9

packages/indexer-api/package.json
james-a-morris: cc @melisaguevara & @daywiss  - I couldn't find any major issues when testing but is there special consideration that should be made?

docker/dev.command.sh
james-a-morris: I am thinking about this a bit more. It would make sense to document this better & use a separate ENV var to reference any migrations - this way, we can avoid enforcing running the migrations on specific app names.

docker/prod.command.sh
james-a-morris: https://github.com/across-protocol/indexer/pull/17#discussion_r1736523541

docker/prod.command.sh
james-a-morris: https://github.com/across-protocol/indexer/pull/17#discussion_r1736523541

packages/indexer-api/package.json
james-a-morris: synced offline - @amateima

melisaguevara (approved):

docker/dev.command.sh
amateima: > @amateima do you see any issues if we delete this condition? I think it's ok that we try to run migrations when running any of the apps. If migrations were already executed the command will do nothing.

I think the issue is if the migrations are executed in the same time. One set of migrations are going to fail

amateima: > I am thinking about this a bit more. It would make sense to document this better & use a separate ENV var to reference any migrations - this way, we can avoid enforcing running the migrations on specific app names.

Yeah, we'll discuss this when you are back 👍

amateima (approved):

PR #18, opened 2024-08-26
feat: index spoke pool events

melisaguevara:
This PR includes:
- new Database entities for: ExecutedRelayerRefundRoot, RelayedRootBundle and TokensBridged events emitted by the SpokePool contracts.
- Logic for extracting those events from the SpokePoolClient and save them in the postgres database.

To do:
- RequestedSpeedUpV3 events are missing because they are not directly exposed by the SDK. They will be included when [this PR](https://github.com/across-protocol/sdk/pull/713) of the SDK is merged.


packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -0,0 +1,193 @@
...
+        "There was an error while saving RelayedRootBundle events:",
+        error,
+      );
+    }
+  }
+
+  public async formatAndSaveExecutedRelayerRefundRootEvents(
+    executedRelayerRefundRootEvents: (across.interfaces.RelayerRefundExecutionWithBlock & {
+      caller: string;
+    })[],
```
melisaguevara, line 142: I'll modify this type definition when the version of the SDK is bumped including [this PR](https://github.com/across-protocol/sdk/pull/713).

packages/indexer-database/src/entities/evm/TokensBridged.ts
```diff
@@ -0,0 +1,49 @@
+import {
+  Column,
+  CreateDateColumn,
+  Entity,
+  PrimaryGeneratedColumn,
+  Unique,
+} from "typeorm";
+
+@Entity({ schema: "evm" })
+@Unique("UK_tokensBridged_chainId_leafId_l2TokenAddress_transactionHash", [
```
amateima, line 10: No need to change now, but in the future such long indexes can be shortened
e.g `UK_tokensBridged_chain_leaf_l2Token_txHash`

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -0,0 +1,193 @@
...
+    const formattedEvents = v3FundsDepositedEvents.map((event) => {
+      return {
+        ...event,
+        relayHash: getRelayHashFromEvent(event),
+        ...this.formatRelayData(event),
+        quoteTimestamp: new Date(event.quoteTimestamp * 1000),
+      };
+    });
+    try {
+      await v3FundsDepositedRepository?.save(formattedEvents, { chunk: 2000 });
```
amateima, line 51: Nice. I did not know about the `chunk` parameter

amateima, line 51: I prefer to use the explicit `.insert()` function. `.save()` performs a `find` and then an `insert` / `update` which can end up in racing conditions if 2 saves are performed in parallel. This will lead to data corruption

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -0,0 +1,193 @@
...
+      this.logger.info(
+        `Saved ${v3FundsDepositedEvents.length} V3FundsDeposited events`,
+      );
+    } catch (error) {
+      this.logger.error(
+        "There was an error while saving V3FundsDeposited events:",
+        error,
+      );
+    }
```
amateima, line 60: I don't think the error propagation should be stopped here. it should be up to the caller to decide if it wants to ignore the error or handle it. There's also the option to add an additional function parameter `throwError = true` if the caller doesn't want error propagation (will pass `false` as value)

amateima (commented):
Looks good. I super like the code reduction in the `deposits.ts` service and isolating the database operations in the repository file. Left a few comments. Happy to discuss them if needed 🙏

packages/indexer-database/src/entities/evm/ExecutedRelayerRefundRoot.ts
```diff
@@ -0,0 +1,57 @@
...
+  l2TokenAddress: string;
+
+  @Column()
+  amountToReturn: string;
+
+  @Column({ type: "jsonb" })
+  refundAmounts: string[];
+
+  @Column({ type: "jsonb" })
+  refundAddresses: string[];
```
james-a-morris, line 38: OOC doesn't postgres support array types?

packages/indexer-database/src/entities/evm/RelayedRootBundle.ts
```diff
@@ -0,0 +1,44 @@
...
+  transactionHash: string;
+
+  @Column()
+  transactionIndex: number;
+
+  @Column()
+  logIndex: number;
+
+  @Column()
+  blockNumber: number;
```
james-a-morris, line 40: should we impose any constraints on this given we should always have positive integers?

packages/indexer-database/src/entities/evm/ExecutedRelayerRefundRoot.ts
melisaguevara: Yes it does. @amateima do you have any opinions on wether we should use jsonb or arrays for this kind of fields?
I think it's common to use jsonb but I'm not strongly opinionated.

packages/indexer/src/database/SpokePoolRepository.ts
melisaguevara: I used `.save()` to be able to use the `chunk` parameter as .insert() doesn't support it. But let me see if we can use an `insertQueryBuilder` to keep that functionality.

packages/indexer-database/src/entities/evm/ExecutedRelayerRefundRoot.ts
amateima: jsonb is more than good 👍

packages/indexer/src/database/SpokePoolRepository.ts
amateima: I guess you would have to chunk the list manually. [In the scraper](https://github.com/across-protocol/scraper-api/blob/master/src/utils.ts#L147-L151) we do that a lot

packages/indexer-database/src/entities/evm/TokensBridged.ts
melisaguevara: I modified this one and the one for ExecutedRelayerRefundRoot here [097e5bf](https://github.com/across-protocol/indexer/pull/18/commits/097e5bfbb950550c3446e3382f7a72e4b6c36450)

packages/indexer/src/database/SpokePoolRepository.ts
melisaguevara: Modified here [3cc6cc6](https://github.com/across-protocol/indexer/pull/18/commits/3cc6cc621d84274a01f677a9d7cffdd52951f478)

packages/indexer/src/database/SpokePoolRepository.ts
melisaguevara: I like the option of the `throwError` param. I implemented it here [5f0aaa0](https://github.com/across-protocol/indexer/pull/18/commits/5f0aaa0e0f6e2403d7a64b0740c6a9abab55c775).
I decided to set the default to false for now as every time you run the indexer now you will get duplicated events and the save functions will throw. I think we can change it to true once we work on improving the indexing loop.

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -47,8 +49,13 @@ export class SpokePoolRepository {
         quoteTimestamp: new Date(event.quoteTimestamp * 1000),
       };
     });
+    const chunkedEvents = across.utils.chunk(formattedEvents, this.chunkSize);
     try {
-      await v3FundsDepositedRepository?.save(formattedEvents, { chunk: 2000 });
+      await Promise.all(
+        chunkedEvents.map((eventsChunk) =>
+          v3FundsDepositedRepository?.insert(eventsChunk),
```
amateima, line 56: I think having the repository potentially `undefined` is a bit confusing. 
The repository should carry only about interacting with DB entities/tables and not handling anything related to db connection management. 
What are your thoughts about removing ` | undefined` from the type of the `postgres` constructor parameter? If there is no postgres connection, then we should not have any repository at all

melisaguevara: Yes, I agree. I also noticed this while creating the `HubPoolRepository` and was already planning to change it.
I'll modify it!

melisaguevara: Modified here [15a6453](https://github.com/across-protocol/indexer/pull/18/commits/15a64533c604ef7ecac854f731f602f2c49af727)

daywiss (approved):
great!

amateima (approved):

PR #21, opened 2024-08-30
feat(SpokePool): index RequestedSpeedUpV3Deposit events

melisaguevara:
This PR includes:
- New entity and migration for RequestedSpeedUpV3Deposit events.
- A refactoring of the SpokePoolRepository including the use of BaseRepository.
- A change in the Repository classes: throwError is now a property of BaseRepository but can also be overridden when calling any method of the repositories.


packages/indexer/src/services/deposits.ts
```diff
@@ -293,11 +293,12 @@ export async function Indexer(config: Config) {
       }),
     );
 
+  const dbThrowError = false;
```
amateima, line 296: What's the reason for silencing the errors? Don't we want the events to be scraped again if for exemple there was a problem in the storage layer?

melisaguevara: This is for avoiding errors while developing as we don't have the indexing loop in place yet. It should be temporary, I'll add a TODO comment to update this when we improve the indexing loop.

daywiss (approved):

amateima (approved):

PR #23, opened 2024-09-05
feat: add more robust block range update tracking

daywiss:
## motivation
We want a more robust way to track what block ranges we have sucessfully queried. 

## changes
This changes how we do this by instantiating a single spoke pool client along with its own hub/config store clients, it will now attempt small ranges of blocks and save the result to redis ( success/failure) that way we can either retry or avoid reprocessing block ranges.  This currently only stores spoke pool events. 

## performance
this relies heavily on caching features of the cached rpc client, since there are many redundant queries for hub and config store clients.  both of these clients are re-queried for every spoke pool block range update. if we can optimize this futher that would be great


packages/indexer/src/services/commands.ts
```diff
@@ -0,0 +1,124 @@
...
+  return (command as CommandError).error !== undefined;
+}
+export function isCommandBase(command: Command): command is CommandError {
+  return !isCommandResult(command) && !isCommandError(command);
+}
+type Config = {
+  redis: Redis;
+  prefix: string;
+};
+export class CommandStore {
```
daywiss, line 42: this class is basically just meant to store block ranges that we are either: processing, sucessfully processed, or had an error

packages/indexer/src/services/indexer.ts
```diff
@@ -0,0 +1,479 @@
...
+    maxBlockLookBack: number;
+  };
+  spokeConfig: {
+    chainId: number;
+    providerUrl: string;
+    maxBlockLookBack: number;
+  };
+  redisKeyPrefix: string;
+};
+export function Indexer(config: Config) {
```
daywiss, line 204: this service has a lot in common with the original deposit indexer service, lots of code copied and pasted here, eventually this would replace that service, since we have the ability to monitor which block ranges we have successfully stored

packages/indexer/src/services/indexer.ts
```diff
@@ -0,0 +1,479 @@
...
+    ...retryProviderConfig,
+    cache: redisCache,
+    logger,
+    ...spokeConfig,
+  });
+
+  const spokePoolClientRepository = new SpokePoolRepository(postgres, logger);
+  const hubPoolRepository = new HubPoolRepository(postgres, logger);
+
+  async function tick() {
```
daywiss, line 255: this is the main logical function, should automatically update ranges, store events and possibily repair broken queries ( not implemented )

packages/indexer/src/services/indexer.ts
```diff
@@ -0,0 +1,479 @@
...
+            error,
+          });
+        } else {
+          // not an error type, throw it and crash app likely
+          throw error;
+        }
+      }
+    }
+  }
+  async function updateUnprocessedRanges(toBlock?: number) {
```
daywiss, line 302: this funciton is still a WIP, but the idea is that we have a collection of block ranges that have succeeded, failed, or are in progress.  This should compare that to the latest block, and determine what ranges we need to queue up to process,

packages/indexer/src/services/indexer.ts
```diff
@@ -0,0 +1,479 @@
...
+      maxBlockLookBack: spokeConfig.maxBlockLookBack,
+    });
+
+    for (const range of paginatedBlockRanges) {
+      allPendingCommands.push(await addRange(...range));
+    }
+    return allPendingCommands;
+  }
+
+  async function fetchEventsByRange(fromBlock: number, toBlock: number) {
```
daywiss, line 339: This is the main event fetching logic, initializes all clients, updates them then returns relevent events we want to store

packages/indexer/src/services/indexer.ts
```diff
@@ -0,0 +1,479 @@
...
+    return {
+      v3FundsDepositedEvents,
+      filledV3RelayEvents,
+      requestedV3SlowFillEvents,
+      relayedRootBundleEvents,
+      executedRelayerRefundRootEvents,
+      tokensBridgedEvents,
+    };
+  }
+  async function storeEvents(params: {
```
daywiss, line 422: takes a map of all the events and stores them in postgres

packages/indexer/src/services/indexer.ts
```diff
@@ -0,0 +1,479 @@
...
+      spokeConfig.chainId,
+    );
+    await spokePoolClientRepository.formatAndSaveExecutedRelayerRefundRootEvents(
+      executedRelayerRefundRootEvents,
+    );
+    await spokePoolClientRepository.formatAndSaveTokensBridgedEvents(
+      tokensBridgedEvents,
+    );
+  }
+  async function addRange(fromBlock: number, toBlock: number) {
```
daywiss, line 458: helper function that lets us persist a range query to redis, will error if we are saving one we have already enqueued.  does not check error/completed range queries.

packages/indexer/src/services/indexer.ts
```diff
@@ -0,0 +1,479 @@
...
+    redisKeyPrefix,
+    hubConfig,
+    spokeConfig,
+    configStoreConfig,
+  } = config;
+
+  function makeId(...args: Array<string | number>) {
+    return [redisKeyPrefix, ...args].join(":");
+  }
+  const pendingCommands = new commands.CommandStore({
```
daywiss, line 219: these "commands" just represent block range queries in a certain state: pending, resolved, rejected.
right now we manually just delete and set ranges between each of these tables, not acid compliant, but its ok for a proof of concept. would only really fail if we ended the app right between a delete and set, or if one or the other fails

packages/indexer/src/services/indexer.ts
```diff
@@ -0,0 +1,479 @@
...
+    logger,
+    ...hubConfig,
+  });
+  const configStoreProvider = getRetryProvider({
+    ...retryProviderConfig,
+    cache: redisCache,
+    logger,
+    ...configStoreConfig,
+  });
+  const spokePoolProvider = getRetryProvider({
```
daywiss, line 245: note we only intanciate a single spoke pool for this service, we will need one service for each spoke pool we want to index

packages/indexer/src/services/commands.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+    await this.store.set(this.id(command.id), update);
+    return update;
+  }
+  async has(id: string) {
+    return !!(await this.store.get(this.id(id)));
+  }
+  async setResult(id: string, result: JSON) {
+    const cmd = await this.get(id);
+    assert(isCommandBase(cmd), "Command already has result or error");
```
melisaguevara, line 78: should this check if command is CommandError or CommandResult instead of CommandBase?

daywiss: removed this function for now!

packages/indexer/src/services/spokePoolEvents.ts
```diff
@@ -0,0 +1,432 @@
...
+    maxBlockLookBack: number;
+  };
+  spokeConfig: {
+    chainId: number;
+    providerUrl: string;
+    maxBlockLookBack: number;
+  };
+  redisKeyPrefix: string;
+};
+export async function SpokePoolEvents(config: Config) {
```
melisaguevara, line 200: Would it make sense to turn this into a class? Or do you think it's better to keep this as a function?

packages/indexer/src/services/spokePoolEvents.ts
```diff
@@ -0,0 +1,432 @@
...
+    params.maxConcurrency,
+    params.providerCacheNamespace,
+    params.pctRpcCallsLogged,
+    params.cache,
+    params.standardTtlBlockDistance,
+    params.noTtlBlockDistance,
+    params.providerCacheTtl,
+    params.logger,
+  );
+}
```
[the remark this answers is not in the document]
daywiss: yes, let me save that for future pr when i add that feature!

melisaguevara (approved):
lgtm!

packages/indexer/src/services/spokePoolEvents.ts
james-a-morris: I'm not sure I follow. It looks like we can have a constructor reference the initialization logic (216-259), scope the `tick` function as public, and make the rest private.

daywiss: there is an async call to fetch the top level hub and config clients: https://github.com/across-protocol/indexer/pull/23/files#diff-7b1a9cecc479c0cbfba7d248647a84a648625851bc26ef915a6c821b9f507631R253
which cannot be included in a constructor.   this could be moved to a secondary async initialization funciton, or do some kind of singleton init in the tick function, but that seemed overly complex.  but im happy to massage this into a class if that is what people prefer!

pnpm-lock.yaml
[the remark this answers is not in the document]
daywiss: i dug into this. turns out master has a bad lockfile installed, when i installed lodash this triggered pnpm to repair the file and that is the source of all the changes. not sure how we can avoid this in the future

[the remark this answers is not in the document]
daywiss: if you checkout master, run `pnpm install` in root, im seeing this error: `WARN  Ignoring broken lockfile at /home/dev/src/risklabs/indexer: Lockfile /home/dev/src/risklabs/indexer/pnpm-lock.yaml`. see if you get something similar

packages/indexer/src/services/spokePoolEvents.ts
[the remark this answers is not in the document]
daywiss: i made this change in the hubpool event indexer pr coming up next

packages/indexer/src/main.ts
```diff
@@ -124,33 +125,69 @@ export async function Main(
...
+  // });
+  // const depositIndexer = await services.deposits.Indexer({
+  //   spokePoolProviderUrls,
+  //   hubPoolProviderUrl,
+  //   logger,
+  //   redis,
+  //   postgres,
+  //   indexerQueuesService,
+  //   retryProviderConfig,
+  // });
```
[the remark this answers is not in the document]
daywiss: i can just remove this for now and prefer the code in this pr for now. will fast follow with rest of features

packages/indexer/src/services/spokePoolEvents.ts
```diff
@@ -0,0 +1,432 @@
...
+
+export async function getSpokeClient(
+  params: GetSpokeClientParams,
+): Promise<across.clients.SpokePoolClient> {
+  const { provider, logger, maxBlockLookBack, chainId, hubPoolClient } = params;
+  const address = getDeployedAddress("SpokePool", chainId);
+  const deployedBlockNumber = getDeployedBlockNumber("SpokePool", chainId);
+
+  const toBlock = params.toBlock ?? (await provider.getBlockNumber());
+  const fromBlock = params.fromBlock ?? deployedBlockNumber;
```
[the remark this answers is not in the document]
daywiss: redis is used to manage block ranges external to the client, see here where we fetch known completed ranges: https://github.com/across-protocol/indexer/pull/23/files#diff-7b1a9cecc479c0cbfba7d248647a84a648625851bc26ef915a6c821b9f507631R310

packages/indexer/src/services/spokePoolEvents.ts
```diff
@@ -0,0 +1,432 @@
...
+    address,
+    deployedBlockNumber,
+    ...eventSearchConfig,
+    blockRangeSearched: toBlock - fromBlock,
+  });
+  const spokePoolContract = new Contract(
+    address,
+    SpokePoolFactory.abi,
+    provider,
+  );
```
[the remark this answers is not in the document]
daywiss: i can give it a shot

packages/indexer/src/rangeQueryStore.ts
```diff
@@ -0,0 +1,63 @@
...
+export type RangeQuery = {
+  id: string;
+  fromBlock: number;
+  toBlock: number;
+};
+export type Range = [number, number];
+export type Ranges = Range[];
+
+// Uses redis set to store any range which is just a from and to block
+export class RangeQueryStore {
```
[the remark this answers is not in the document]
daywiss: not sure what you mean, theres a few private functions there, the rest should be public

packages/indexer/src/main.ts
```diff
@@ -124,33 +125,69 @@ export async function Main(
...
+        hubConfig: {
+          chainId: hubPoolNetworkInfo.chainId,
+          providerUrl: hubPoolProviderUrl,
+          maxBlockLookBack: 10000,
+        },
+        spokeConfig: {
+          chainId: spokePoolNetworkInfo.chainId,
+          providerUrl: spokePoolProviderUrl,
+          // TODO: Set this per chain
+          maxBlockLookBack: 10000,
```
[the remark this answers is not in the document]
daywiss: is there a way to include specific look back per chain, or are you saying i should copy them into a constants file?

packages/indexer/src/main.ts
```diff
@@ -124,33 +125,69 @@ export async function Main(
...
+        },
+        redisKeyPrefix: `spokePoolIndexer:${spokePoolNetworkInfo.chainId}`,
+      };
```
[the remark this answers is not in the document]
daywiss: i can try to clean it up a bit

packages/indexer/src/services/spokePoolEvents.ts
```diff
@@ -0,0 +1,429 @@
...
+            errorMessage: error.message,
+          });
+        } else {
+          // not an error type, throw it and crash app likely
+          throw error;
+        }
+      }
+    }
+  }
+  async function getUnprocessedRanges(toBlock?: number): Promise<Ranges> {
```
amateima, line 293: This function assumes that `blockLookBack` is the same during different launches of the indexer, right? If I start the indexer with a lookback of 10k, stop it, and then restarted with a lookback 100k, all the blocks will be processed again

packages/indexer/src/main.ts
```diff
@@ -136,33 +178,32 @@ export async function Main(
...
+        message: "Starting indexer",
+        ...config,
+      });
+      const spokeIndexer = await services.spokePoolEvents.SpokePoolEvents({
+        logger,
+        redis,
+        postgres,
+        ...config,
+      });
+      await spokeIndexer.tick();
```
amateima, line 202: Can we add polling logic here? Query events every x seconds. This is how the indexer is supposed to run

packages/indexer/src/main.ts
```diff
@@ -136,33 +178,32 @@ export async function Main(
...
-  });
-
-  // TODO: add looping to keep process going
-  // do {
-  logger.info("index loop starting");
-  await depositIndexer(Date.now());
-  logger.info("index loop complete");
-  // sleep(30000);
-  // } while (running);
+  if (postgres && redis && retryProviderConfig) {
```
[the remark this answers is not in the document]
daywiss: previously these were optional, we can just make them required at this point

packages/indexer/src/rangeQueryStore.ts
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * Creates a new block range entry in Redis.
+   * @param {number} fromBlock - The starting block of the range.
+   * @param {number} toBlock - The ending block of the range.
+   * @throws Will throw an error if the range already exists.
+   * @returns {Promise<RangeQuery>} The created range query.
+   */
+  async create(fromBlock: number, toBlock: number) {
+    const id = this.id(fromBlock, toBlock);
+    assert(!(await this.has(id)), "Range already exists");
```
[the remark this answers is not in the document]
daywiss: removing unused functions

packages/indexer/src/rangeQueryStore.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+  /**
+   * Adds a range entry to Redis by specifying the block range.
+   * @param {number} fromBlock - The starting block of the range.
+   * @param {number} toBlock - The ending block of the range.
+   * @returns {Promise<RangeQuery>} The range query associated with the id.
+   */
+  async setByRange(fromBlock: number, toBlock: number) {
+    return this.set(this.id(fromBlock, toBlock));
+  }
```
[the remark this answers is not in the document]
daywiss: yes will remove unused functions

packages/indexer/src/rangeQueryStore.ts
```diff
@@ -0,0 +1,129 @@
+import assert from "assert";
```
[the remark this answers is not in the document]
daywiss: goign to move this to a redis folder since database implies postgres classes

packages/indexer/src/services/spokePoolEvents.ts
```diff
@@ -0,0 +1,429 @@
...
+    maxBlockLookBack: number;
+  };
+  spokeConfig: {
+    chainId: number;
+    providerUrl: string;
+    maxBlockLookBack: number;
+  };
+  redisKeyPrefix: string;
+};
+export async function SpokePoolEvents(config: Config) {
```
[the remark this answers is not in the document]
daywiss: will rename this accordingly

packages/indexer/src/services/spokePoolEvents.ts
daywiss: yea good point, it does use block range pairs, so changes to the lookback will cause it to miss and rerun

packages/indexer/src/services/spokePoolEvents.ts
amateima: For me it's ok to stick with this implementation

packages/indexer/src/main.ts
```diff
@@ -189,9 +188,30 @@ export async function Main(
...
+  do {
+    logger.info({
+      message: "Running indexers",
+    });
+    await Promise.all(spokePoolIndexers.map((s) => s.tick()));
+    await across.utils.delay(10);
+    logger.info({
+      message: "Completed running indexers",
+    });
+  } while (!exitRequested);
```
amateima, line 213: Doing the polling logic here breaks the decoupling between the spoke pool clients as all clients need to finish in order to move to the next block range. The clients will need different polling intervals depending on the block time. 
I think the polling logic should be part of the `SpokePoolIndexer` as the main.ts entrypoint would not have to deal with such an implementation detail 

```
SpokePoolIndexer {
  start() {
    looping logic {
      tick()
    }
  }
  
  tick() { ... }
}

spokeIndexer.start()
```

amateima (changes requested):

packages/indexer/src/main.ts
daywiss: updated this

packages/indexer/src/rangeQueryStore.ts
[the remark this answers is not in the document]
james-a-morris: I'm of two minds here. I think JSDoc is a low-lift that goes a long way toward maintainability - and it's more convenient to write them when the code is fresh. 

It will also increase overall code readability for external projects that look at this repo to understand what is happening under the hood.

packages/indexer/src/rangeQueryStore.ts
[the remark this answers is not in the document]
james-a-morris: I.e. `public`/`private` when declaring functions.

amateima (approved):

packages/indexer/src/main.ts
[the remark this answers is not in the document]
daywiss: i see, ok yes i still have an issue to refactor configs, so if you dont mind id like to do it in that pr

packages/indexer/src/rangeQueryStore.ts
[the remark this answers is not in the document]
daywiss: explicitly declaring public vs just using the default scope? yea im ok with either way, i can explicitly declare it

PR #29, opened 2024-09-12
feat: add hubpool event saving with block range checks

daywiss:
# motivation
Add hubpool indexer with block range success status in the same way as previous spoke pool indexer.

# changes
This adds hubpoolindexer file which is basically a copy of the spoke pool indexer file but for hub events.  also adds similar config and startup routine as the spoke pool indexer

# todo
future prs will update spoke pool indexer to catch up with what was done in services/deposits.ts and deal with situations where redis and postres are not synced correctly

### DIFF
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -124,7 +124,7 @@ async function getSpokePoolIndexerConfig(params: {
     redisKeyPrefix: `spokePoolIndexer:${spokePoolNetworkInfo.chainId}`,
   };
 }
-// utility call to create the spoke pool event indexer config
+// utility call to create the hubpool event indexer config
 async function getHubPoolIndexerConfig(params: {
   retryProviderConfig: services.deposits.RetryProviderConfig;
   hubPoolNetworkInfo: providers.Network;
@@ -236,7 +236,10 @@ export async function Main(
       hubPoolIndexer.stop();
     } else {
       logger.info("\nForcing exit...");
-      across.utils.delay(1).finally(() => process.exit());
+      redis?.quit();
+      postgres?.destroy();
+      logger.info("Exiting indexer");
+      across.utils.delay(5).finally(() => process.exit());
     }
   });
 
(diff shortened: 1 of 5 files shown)


packages/indexer/src/main.ts
```diff
@@ -211,6 +244,7 @@ export async function Main(
     message: "Running indexers",
     at: "Indexer#Main",
   });
+  await hubPoolIndexer.start(10);
```
james-a-morris, line 247: Can we put this on line 249? 
```
const [bundleResults, hubResults, ...spokeResults] ...
```

packages/indexer/src/main.ts
```diff
@@ -124,6 +124,24 @@ async function getSpokePoolIndexerConfig(params: {
...
+  const { retryProviderConfig, hubPoolProviderUrl, hubPoolNetworkInfo } =
+    params;
+  return {
+    retryProviderConfig,
+    hubConfig: {
+      chainId: hubPoolNetworkInfo.chainId,
+      providerUrl: hubPoolProviderUrl,
+      maxBlockLookBack: 10000,
+    },
+    redisKeyPrefix: `hubPoolIndexer:${hubPoolNetworkInfo.chainId}`,
```
james-a-morris, line 142: Nit: (and probably not actionable) but this should only really be 1 or 11155111

packages/indexer/src/services/hubPoolIndexer.ts
```diff
@@ -0,0 +1,221 @@
...
+    const allQueries = await resolvedRangeStore.entries();
+    const resolvedRanges = allQueries.map(([, x]) => [x.fromBlock, x.toBlock]);
+    const needsProcessing = differenceWith(
+      allPaginatedBlockRanges,
+      resolvedRanges,
+      isEqual,
+    );
+
+    logger.info({
+      message: `${needsProcessing.length} block ranges need processing`,
```
james-a-morris, line 135: Nit: can we include an `at`? It may help the logs if we need to back out where a function is

packages/indexer/src/services/hubPoolIndexer.ts
```diff
@@ -0,0 +1,221 @@
...
+      message: "updating hubpool client",
+      fromBlock,
+      toBlock,
+    });
+    await hubPoolClient.update();
+    logger.info({
+      message: "updated hubpool client",
+      fromBlock,
+      toBlock,
+    });
```
james-a-morris, line 166: The hub's update function should do logging like this already

amateima (approved):

melisaguevara (approved):

amateima:
> deal with situations where redis and postres are not synced correctly

I don't think that's necessary as this implementation will change will the re-org mitigation implementation

PR #33, opened 2024-09-13
improve: refactor into OOP

james-a-morris:
WIP change to move our indexers over to using OOP. We have certain patterns in our current structure that could be condensed via shared code from inheritence


packages/indexer/src/main.ts
```diff
@@ -252,6 +252,7 @@ export async function Main(
     await Promise.allSettled([
       bundleProcessor(),
       hubPoolIndexer.start(10),
+      Promise.resolve(),
```
melisaguevara, line 255: Can this line be removed?

packages/indexer/src/services/hubPoolIndexer.ts
```diff
@@ -23,90 +22,94 @@ type Config = {
...
-  } = config;
-
-  let stopRequested = false;
 
-  function makeId(...args: Array<string | number>) {
-    return [redisKeyPrefix, ...args].join(":");
+/**
+ * Indexer for the hubpool contract and its component events
+ */
+export class Indexer extends BaseIndexer {
```
melisaguevara, line 29: Should we keep the name as HubPoolIndexer?

melisaguevara (commented):
Looks really nice! Just two comments

packages/indexer/src/services/hubPoolIndexer.ts
daywiss: it could make sense to name the main classes "Indexer", then you just references `serivces.hubpoolIndexer.Indexer` or `services.spokePoolIndexer.Indexer`, so i am probably ok his original naming

daywiss (approved):

amateima (approved):

PR #36, opened 2024-09-17
Update exec files

evaldofelipe:


james-a-morris (commented):
The diff is nil

amateima (approved):

amateima:
> The diff is nil

It was just about changing file permissions

PR #37, opened 2024-09-17
feat: spokePoolProcessor

melisaguevara:
This PR includes a new processor for SpokePool events. It runs after every update of a spoke pool client and:
- Updates relayHashInfo table based on new events.
- Updates the status of relayHashInfo rows based on the associated events.
- Tries to identify refunds for expired deposits based on ExecutedRelayerRefundRoot events.


packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -0,0 +1,232 @@
...
+      typeof entities.RequestedV3SlowFill,
+      string,
+    ],
+  ]) {
+    const eventsWithoutRelayHashInfo = await postgres
+      .getRepository(eventEntity)
+      .createQueryBuilder("events")
+      .leftJoin("relay_hash_info", "rhi", `events.id = rhi.${fieldName}`)
+      .where(`rhi.${fieldName} IS NULL`)
+      .getMany();
```
amateima, line 57: instead of joining tables having millions of rows, can we work only with the events returned by the contract clients by passing them as a parameter? I think we might want to limit this process step to only the events returned by the `update()` function. We can discuss this individually for each function out of `assignSpokeEventsToRelayHashInfo`, `updateRelayStatus`, `updateRefundedDepositsStatus`

amateima (changes requested):

packages/indexer/src/services/spokePoolProcessor.ts
melisaguevara: I think we could use that approach for updating `relayHashInfo` event rows and potentially for refunds too.
Using it also for `updateRelayStatus` will make us lose track of expired deposits.
One question I have is related to error handling, if any of those steps throws, should we let the indexer break? or should we somehow track the failed updates to retry them later?

amateima (approved):

PR #38, opened 2024-09-17
feat(bundles): add block range tracking

james-a-morris:
We should track start/end block ranges into their own entity so that resolving a bundles range requires a single lookup.

### DIFF
--- a/packages/indexer/src/database/BundleRepository.ts
+++ b/packages/indexer/src/database/BundleRepository.ts
@@ -109,6 +109,8 @@ export class BundleRepository extends utils.BaseRepository {
       .leftJoin("bundle_block_range", "br", "b.id = br.bundleId")
       .where("br.bundleId IS NULL")
       .select(["b.id", "proposal"])
+      .orderBy("proposal.blockNumber", "ASC")
+      .limit(1000) // Limit to 1000 bundles to process at a time
       .getMany();
   }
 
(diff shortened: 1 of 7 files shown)


amateima (approved):

james-a-morris:
cc @amateima RE: https://github.com/across-protocol/indexer/pull/38/commits/99a2f66639bb55862f49544e3ec5126f3f67c17d & https://github.com/across-protocol/indexer/pull/38/commits/2fb5c7cbd48ca69cb36a5cd456f7ec96009dc5eb

I've updated some comments, added 3 hours to the lookback on disputes to account for early dispute windows being longer than an hour, and set the `maxLookbackFromBlock` to explicitly search the entire range if not defined.

packages/indexer/src/database/BundleRepository.ts
```diff
@@ -93,27 +99,40 @@ export class BundleRepository extends utils.BaseRepository {
...
+  > {
+    return this.postgres
+      .getRepository(entities.Bundle)
+      .createQueryBuilder("b")
+      .leftJoinAndSelect("b.proposal", "proposal")
+      .leftJoin("bundle_block_range", "br", "b.id = br.bundleId")
+      .where("br.bundleId IS NULL")
+      .select(["b.id", "proposal"])
+      .orderBy("proposal.blockNumber", "ASC")
+      .limit(1000) // Limit to 1000 bundles to process at a time
```
james-a-morris, line 113: @amateima setting a cap of 1000 so that we can slowly populate the backfill

daywiss (approved):

amateima (approved):

melisaguevara (approved):

PR #41, opened 2024-09-18
fix: simplify docker setup

amateima:
Because of some issues that @evaldofelipe encountered while setting the CI/CD flow, I simplified the Docker setup.

- Dockerfile contains a single stage image now
- `node` app now has a `start:prod` command. `ts-node` is not recommended to be used in production
- the root package.json has new commands for running the apps in production: `start:indexer:prod` and `start:indexer-api:prod`


Dockerfile
```diff
@@ -1,52 +1,18 @@
...
-
-COPY apps/node/package.json ./apps/node/package.json
-COPY packages/indexer/package.json ./packages/indexer/package.json
-COPY packages/indexer-api/package.json ./packages/indexer-api/package.json
-COPY packages/indexer-database/package.json ./packages/indexer-database/package.json
-COPY packages/persistence-example/package.json ./packages/persistence-example/package.json
-COPY packages/template/package.json ./packages/template/package.json
-COPY packages/typescript-config/package.json ./packages/typescript-config/package.json
+# Copy the rest of the files & build the app
+COPY . .
```
james-a-morris, line 11: This will trigger an install on any change to the repo. I think this will make the DX slower to run

docker-compose.yml
```diff
@@ -25,17 +25,14 @@ services:
   indexer-scraper:
     build:
       context: .
-      target: development
     container_name: indexer_service
     volumes:
-      - .:/usr/src/app
+      # - .:/usr/src/app
```
james-a-morris, line 30: Is this meant to be commented?

amateima: Yeah, mounting the local files in the container led to having some weird package conflicts. I commented this until I figure out exactly what's happening

james-a-morris (approved):

PR #45, opened 2024-09-24
improve(indexer): refactor configuration parsing

daywiss:
# Motivation
We wanted to simplify infrastructure env management as well as cleanup the parsing logic in the indexer.

# Changes
This makes several changes:
1. Moves a lot of the logic around ENV parsing to a utlity file, to clean up the main file
2. Exports a single `envToConfig` function, which parses env before getting sent to main function, allowing better intellisense on configuration
3. Providers can now be specified as a list, using a comma separated format: `RPC_PROVIDER_URLS_137=url1,url2`
4. You can now enable spoke pools through a new env `SPOKEPOOL_CHAINS_ENABLED=1,10`
5. You must a hubpool chain network `HUBPOOL_CHAIN=1`
6. Some minor configuration changes to the indexers to simplify and remove some redundancy
7. added env to readme, though this may change in future to be simpler

# Todo
1. Future pr will try to use this same pattern to the indexer-api package
2. Should add defaults to replace some of the complexity specifying retry provider envs
3. env for maxBlockLookback should probably also be defaulted, and/or specified in env

### DIFF
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -1,86 +1,9 @@
-import assert from "assert";
 import * as services from "./services";
 import winston from "winston";
 import Redis from "ioredis";
 import * as across from "@across-protocol/sdk";
 import { connectToDatabase } from "./database/database.provider";
-import { DatabaseConfig } from "@repo/indexer-database";
 import * as utils from "./utils";
-
-type Config = {
-  redisConfig: utils.RedisConfig;
-  postgresConfig: DatabaseConfig;
-  spokeConfigs: Omit<
-    services.spokePoolIndexer.Config,
-    "logger" | "redis" | "postgres"
-  >[];
-  hubConfig: Omit<
-    services.hubPoolIndexer.Config,
-    "logger" | "redis" | "postgres"
-  >;
-};
-
-export function envToConfig(env: utils.Env): Config {
-  assert(env.HUBPOOL_CHAIN, "Requires HUBPOOL_CHAIN");
-  const redisConfig = utils.parseRedisConfig(env);
-  const postgresConfig = utils.parsePostgresConfig(env);
-  const allProviderConfigs = utils.parseProviderConfigs(env);
-  const retryProviderConfig = utils.parseRetryProviderConfig(env);
-  const hubPoolChain = utils.parseNumber(env.HUBPOOL_CHAIN);
-  const spokePoolChainsEnabled = utils
-    .parseArray(env.SPOKEPOOL_CHAINS_ENABLED)
-    .map(utils.parseNumber);
-  const providerConfigs = allProviderConfigs.filter(
-    (provider) => provider[1] === hubPoolChain,
-  );
-  assert(
-    allProviderConfigs.length > 0,
-    `Requires at least one RPC_PROVIDER_URLS_CHAINID`,
-  );
-
-  const hubConfig = {
-    retryProviderConfig: {
-      ...retryProviderConfig,
-      chainId: hubPoolChain,
-      providerConfigs,
-    },
-    hubConfig: {
-      chainId: hubPoolChain,
-      maxBlockLookBack: 10000,
-    },
-    redisKeyPrefix: `hubPoolIndexer:${hubPoolChain}`,
-  };
-
-  const spokeConfigs = spokePoolChainsEnabled.map((chainId) => {
-    const providerConfigs = allProviderConfigs.filter(
-      (provider) => provider[1] == chainId,
-    );
-    assert(
-      providerConfigs.length > 0,
-      `SPOKEPOOL_CHAINS_ENABLED=${chainId} but did not find any corresponding RPC_PROVIDER_URLS_${chainId}`,
-    );
-    return {
-      retryProviderConfig: {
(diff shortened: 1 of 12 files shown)


packages/indexer/src/main.ts
```diff
@@ -2,226 +2,131 @@ import assert from "assert";
...
-const stringToInt = s.coerce(s.number(), s.string(), (value) =>
-  parseInt(value),
-);
-function getRetryProviderConfig(
-  env: Record<string, string | undefined>,
-): RetryProviderConfig {
-  assert(env.PROVIDER_CACHE_NAMESPACE, "requires PROVIDER_CACHE_NAMESPACE");
-  assert(env.MAX_CONCURRENCY, "requires MAX_CONCURRENCY");
-  assert(env.PCT_RPC_CALLS_LOGGED, "requires PCT_RPC_CALLS_LOGGED");
+export function envToConfig(env: utils.Env): Config {
```
amateima, line 23: I like this idea of parsing the env into a typed config object. Should we put this (and the Config type definition) in the same place where all the envUtils sit? I think it would be good to have all config logic in a single place: types, assertions, constructing the global config object, etc

amateima: fyi, I am currently working on the Indexers that have data finalisation built in and I am also thinking of separating the indexers configs from the global config

packages/indexer/src/utils/envUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+      if (chainId && value) {
+        const providerUrls = parseArray(value);
+        providerUrls.forEach((url) => {
+          results.push([url, chainId]);
+        });
+      }
+    }
+  }
+  return results;
+}
```
amateima, line 70: I think it would be best to have an unified approach with the bots when it is about handling providers.

```
RPC_PROVIDER_KEY_INFURA=
RPC_PROVIDER_KEY_ALCHEMY=
// QuickNode providers URLs are different for each chain, so no unique API key
RPC_PROVIDER_QUICKNODE_1=
RPC_PROVIDERS=INFURA,ALCHEMY,QUICKNODE
RPC_PROVIDERS_<OVERWRITE_CHAIN_ID>=ALCHEMY,QUICKNODE,INFURA
```

The outcome of this is that when we want to swap RPCs order, we'll have to swap provider names, not long URLs.

Handling providers env is already part of the bots, so maybe we can just copy some code from there.

@melisaguevara @james-a-morris I am curious to know your input on this. Should we have simple envs such as `RPC_PROVIDERS_1` with the value being a list of RPC urls **or** should we move with the more complex implementation used by the bots?

We can also start with the simpler implementation and make it more complex in the future

packages/indexer/src/utils/envUtils.ts
```diff
@@ -0,0 +1,102 @@
+import assert from "assert";
```
amateima, line 1: I think the logic in this file grew in time and now it has a well defined scope and it deserves to have its own place in the code structure, not just being placed in the generic `utils` directory. 
My suggestion is to place this logic in a `config` directory right under `src`. And there we can mirror the structure I previously proposed in the Tech Spec doc:
- all the models and types placed in a model file
- the envs parsing, validations, assertions and the construction the global config object placed in a service file (or multiple)

packages/indexer/src/main.ts
daywiss: sure it can be moved to the env utils

packages/indexer/src/utils/envUtils.ts
daywiss: imo when i looked at the bot envs it did not seem easy to understand or reason about.  but perhaps im biased towards more explicit env definitions

packages/indexer/src/utils/envUtils.ts
amateima: > i dont think we need a new folder if this is going to be the only file of its kind

Sounds good, we can keep only this file, but I still feel like this should sit higher in the directories tree, as a child of the `src`. 
I agree most of the functions in the file are pure functions, but I see some different nuances there. `parseArray` is like a pure utility function, but others such as `parseRedisConfig`, `parsePostgresConfig`, etc are actually managing the configuration values and this is one of the core pieces of the indexer.

amateima (approved):
This looks good on my end! There's only one active thread which I replied to, but beside that I don't have other comments 👍 

I

packages/indexer/src/utils/envUtils.ts
daywiss: i can make that change

packages/indexer/src/utils/envUtils.ts
daywiss: > We also have the project to move most of the configs to a common repo. If we are going to use almost the same settings that the bots use, I would say it makes sense to use the same approach as them.

yea, i think we can work towards that. it really depends on what use case we are optimizing for, but i think the code will be pretty complex to implement.  so id like to make sure we really want it and do as a separate pr!

packages/indexer/src/utils/envUtils.ts
daywiss: moved it to parseEnv.ts as a top level file and exported it

PR #46, opened 2024-09-24
feat: deposit status endpoint and api refactor

melisaguevara:
This PR includes:
- A refactor of the API structure. It separates the api components into:
  - Router: where the endpoints and its handling function are defined.
  - Controller: where request and response are handled.
  - Service: where the logic for each endpoint is defined.
- The implementation of /deposit/status endpoint.


packages/indexer-api/src/controllers/indexer.ts
```diff
@@ -0,0 +1,114 @@
...
+  public service: IndexerService;
+  constructor(private db: DataSource) {
+    this.service = new IndexerService(this.db);
+  }
+
+  public getDeposits = async (
+    req: Request,
+    res: Response,
+    next: NextFunction,
+  ) => {
```
daywiss, line 72: good use of `= async ()=>` so you auto bind `.this`

amateima (approved):
Looks good!

PR #47, opened 2024-09-24
feat: track spoke pool balances

james-a-morris:
Waiting on PR from @daywiss RE: redis

### DIFF
--- a/packages/indexer/src/database/BundleRepository.ts
+++ b/packages/indexer/src/database/BundleRepository.ts
@@ -134,7 +134,8 @@ export class BundleRepository extends utils.BaseRepository {
 
   /**
    * Retrieves all root bundle executed events that haven't been associated to a bundle relation yet.
-   * @returns A list of partial executed root bundle entities that include the block number, transaction hash, log index, and transaction index
+   * @returns A list of partial executed root bundle entities that include the block number,
+   *          transaction hash, log index, and transaction index
    */
   public retrieveUnassociatedRootBundleExecutedEvents(): Promise<
     PickRangeType<entities.RootBundleExecuted>[]
@@ -168,7 +169,7 @@ export class BundleRepository extends utils.BaseRepository {
    * @param maxLookbackFromBlock The maximum number of blocks to look back from the provided block number (optional)
    * @returns The closest proposed (undisputed/non-canceled) root bundle back in time, or undefined if none are found
    */
-  public retrieveClosestProposedRootBundle(
+  public retrieveClosestProposedRootBundleEvent(
     blockNumber: number,
     transactionIndex: number,
     logIndex: number,
@@ -202,6 +203,32 @@ export class BundleRepository extends utils.BaseRepository {
       )
       .orderBy("prb.blockNumber", "DESC") // Grab the most recent proposal
       .getOne();
+  }
+
+  /**
+   * Retrieves the most recent bundle with the given status. Optionally, the bundle can
+   * be filtered by a block number. Note: this is a stored bundle from the "Bundle" table
+   * that has already been associated with a proposal.
+   * @param status The status of the bundle to retrieve
+   * @param blockNumber The block number to filter by (optional)
+   * @returns The most recent bundle with the given status, or null if none are found
+   */
+  public retrieveMostRecentBundle(
+    status: entities.BundleStatus,
+    blockNumber?: number,
+  ): Promise<entities.Bundle | null> {
+    const queryBuilder = this.postgres
+      .getRepository(entities.Bundle)
+      .createQueryBuilder("b")
+      .leftJoin("b.proposal", "proposal")
+      .where("b.status = :status", { status });
+    if (blockNumber) {
+      queryBuilder.andWhere("proposal.blockNumber < :blockNumber", {
+        blockNumber,
+      });
+    }
+    queryBuilder.orderBy("proposal.blockNumber", "DESC");
+    return queryBuilder.getOne();
   }
 
   /**
--- a/packages/indexer/src/database/BundleRepository.ts
+++ b/packages/indexer/src/database/BundleRepository.ts
@@ -31,7 +31,7 @@ export class BundleRepository extends utils.BaseRepository {
   constructor(
     postgres: DataSource,
     logger: winston.Logger,
-    throwError: boolean,
+    throwError?: boolean,
   ) {
     super(postgres, logger, throwError);
   }
--- a/packages/indexer/src/database/BundleRepository.ts
+++ b/packages/indexer/src/database/BundleRepository.ts
@@ -211,11 +211,13 @@ export class BundleRepository extends utils.BaseRepository {
    * that has already been associated with a proposal.
    * @param status The status of the bundle to retrieve
    * @param blockNumber The block number to filter by (optional)
-   * @returns The most recent bundle with the given status, or null if none are found
+   * @param nthBundle The nth bundle to retrieve (optional) defaults to 0 (most recent)
+   * @returns The most recent nth bundle with the given status, or null if none are found
    */
   public retrieveMostRecentBundle(
     status: entities.BundleStatus,
     blockNumber?: number,
+    nthBundle?: number,
   ): Promise<entities.Bundle | null> {
     const queryBuilder = this.postgres
       .getRepository(entities.Bundle)
@@ -228,6 +230,9 @@ export class BundleRepository extends utils.BaseRepository {
       });
     }
     queryBuilder.orderBy("proposal.blockNumber", "DESC");
+    if (nthBundle !== undefined) {
+      queryBuilder.offset(nthBundle);
+    }
     return queryBuilder.getOne();
   }
 
--- a/packages/indexer/src/database/BundleRepository.ts
+++ b/packages/indexer/src/database/BundleRepository.ts
@@ -222,7 +222,7 @@ export class BundleRepository extends utils.BaseRepository {
     const queryBuilder = this.postgres
       .getRepository(entities.Bundle)
       .createQueryBuilder("b")
-      .leftJoin("b.proposal", "proposal")
+      .leftJoinAndSelect("b.proposal", "proposal")
       .where("b.status = :status", { status });
     if (blockNumber) {
       queryBuilder.andWhere("proposal.blockNumber < :blockNumber", {
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -22,6 +22,7 @@ import {
 } from "./utils/contractFactoryUtils";
 import { SpokePoolIndexerDataHandler } from "./services/SpokePoolIndexerDataHandler";
 import { SpokePoolProcessor } from "./services/spokePoolProcessor";
+import { BundleRepository } from "./database/BundleRepository";
 
 async function initializeRedis(
   config: parseEnv.RedisConfig,
@@ -109,15 +110,16 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
     return spokePoolIndexer;
   });
 
-  const bundleBuilderProcessor = new services.bundleBuilder.Processor({
-    logger,
-    redis,
-    postgres,
-    providerFactory: retryProvidersFactory,
-    hubClientFactory: hubPoolClientFactory,
-    spokePoolClientFactory,
-    configStoreClientFactory,
-  });
+  const bundleBuilderProcessor =
+    new services.bundleBuilder.BundleBuilderService({
+      logger,
+      redis,
+      bundleRepository: new BundleRepository(postgres, logger),
+      providerFactory: retryProvidersFactory,
+      hubClientFactory: hubPoolClientFactory,
+      spokePoolClientFactory,
+      configStoreClientFactory,
+    });
 
   const hubPoolIndexerDataHandler = new HubPoolIndexerDataHandler(
     logger,
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -165,8 +165,8 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
     await Promise.allSettled([
       bundleProcessor.start(10),
       hubPoolIndexer.start(),
-      ...spokePoolIndexers.map((s) => s.start()),
       bundleBuilderProcessor.start(10),
+      ...spokePoolIndexers.map((s) => s.start()),
     ]);
 
   logger.info({
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -116,6 +116,7 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
     providerFactory: retryProvidersFactory,
     hubClientFactory: hubPoolClientFactory,
     spokePoolClientFactory,
+    configStoreClientFactory,
   });
 
   const hubPoolIndexerDataHandler = new HubPoolIndexerDataHandler(
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -114,6 +114,8 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
     redis,
     postgres,
     providerFactory: retryProvidersFactory,
+    hubClientFactory: hubPoolClientFactory,
+    spokePoolClientFactory,
   });
 
   const hubPoolIndexerDataHandler = new HubPoolIndexerDataHandler(
--- a/packages/indexer/src/main.ts
+++ b/packages/indexer/src/main.ts
@@ -159,11 +159,12 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
     at: "Indexer#Main",
   });
   // start all indexers in parallel, will wait for them to complete, but they all loop independently
-  const [bundleResults, hubPoolResult, ...spokeResults] =
+  const [bundleResults, hubPoolResult, bundleBuilderResult, ...spokeResults] =
     await Promise.allSettled([
       bundleProcessor.start(10),
       hubPoolIndexer.start(),
       ...spokePoolIndexers.map((s) => s.start()),
+      bundleBuilderProcessor.start(10),
     ]);
 
   logger.info({
@@ -175,6 +176,8 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
       ),
       bundleProcessorRunSuccess: bundleResults.status === "fulfilled",
       hubPoolIndexerRunSuccess: hubPoolResult.status === "fulfilled",
+      bundleBuilderProcessorRunSuccess:
+        bundleBuilderResult.status === "fulfilled",
     },
   });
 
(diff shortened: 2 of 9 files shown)


daywiss (approved):

packages/indexer/src/database/BundleRepository.ts
```diff
@@ -204,6 +205,37 @@ export class BundleRepository extends utils.BaseRepository {
...
+   */
+  public retrieveMostRecentBundle(
+    status: entities.BundleStatus,
+    blockNumber?: number,
+    nthBundle?: number,
+  ): Promise<entities.Bundle | null> {
+    const queryBuilder = this.postgres
+      .getRepository(entities.Bundle)
+      .createQueryBuilder("b")
+      .leftJoin("b.proposal", "proposal")
```
melisaguevara, line 225: Should we use `.leftJoinAndSelect` here? Because when we use this function we then do `bundle.proposal` but i'm not sure it will be defined if we use only `leftJoin`

packages/indexer/src/main.ts
```diff
@@ -150,11 +161,12 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
     at: "Indexer#Main",
   });
   // start all indexers in parallel, will wait for them to complete, but they all loop independently
-  const [bundleResults, hubPoolResult, ...spokeResults] =
+  const [bundleResults, hubPoolResult, bundleBuilderResult, ...spokeResults] =
```
melisaguevara, line 164: `bundleBuilderResult` should go last to match the `Promise.allSettled` array order

packages/indexer/src/utils/bundleBuilderUtils.ts
```diff
@@ -0,0 +1,124 @@
...
+  current: ProposalRange,
+): ProposalRangeResult[] {
+  return current.chainIds.map((chainId, idx) => ({
+    chainId,
+    startBlock: previous.bundleEvaluationBlockNumbers[idx]
+      ? previous.bundleEvaluationBlockNumbers[idx] + 1
+      : 0, // If this is a new chain, start from block 0
+    endBlock: current.bundleEvaluationBlockNumbers[idx]!,
+  }));
+}
```
melisaguevara, line 81: We have to avoid incrementing the start block for disabled chains.
My proposal is to do something like this:
```
export function getBlockRangeBetweenBundles(
  previous: ProposalRange,
  current: ProposalRange,
): ProposalRangeResult[] {
  return current.chainIds.map((chainId, idx) => {
    const previousStartBlock = previous.bundleEvaluationBlockNumbers[idx] ?? 0;
    const currentEndBlock = current.bundleEvaluationBlockNumbers[idx]!;
    return {
      chainId,
      startBlock:
        previousStartBlock !== currentEndBlock
          ? previousStartBlock + 1
          : previousStartBlock, // Bundle range doesn't change for disabled chains
      endBlock: currentEndBlock,
    };
  });
}
```

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -0,0 +1,235 @@
...
+      bundleDepositsV3,
+      bundleFillsV3,
+      bundleSlowFillsV3,
+      unexecutableSlowFills,
+      expiredDepositsToRefundV3,
+      {
+        hubPoolClient,
+        configStoreClient,
+      },
+    );
```
melisaguevara, line 214: When executing `handleProposedBundleLoop`, should we check that the root we get from the bundleDataClient matches with the one of the proposed bundle before formatting the leaves?

packages/indexer/src/services/bundles.ts
```diff
@@ -31,13 +32,14 @@ class ConfigurationMalformedError extends Error {
...
   private bundleRepository: BundleRepository;
+  private hubPoolRepository: HubPoolRepository;
   constructor(private readonly config: BundleConfig) {
     super(config.logger, "bundle");
   }
 
   protected async indexerLogic(): Promise<void> {
     const { logger } = this.config;
-    const { bundleRepository } = this;
+    const { bundleRepository, hubPoolRepository } = this;
```
melisaguevara, line 42: Is this needed? I don't see any uses for it

packages/indexer/src/utils/bundleBuilderUtils.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (!utils.isDefined(provider)) {
+        throw new Error(`Provider for chain ${chainId} not found`);
+      }
+      if (!utils.isDefined(previousBlock)) {
+        throw new Error(`Previous block number for chain ${chainId} not found`);
+      }
+      const currentBlock = await provider.getBlockNumber();
+      return {
+        chainId,
+        startBlock: previousBlock + 1,
```
melisaguevara, line 110: Same as above, we have to avoid incrementing the start block for disabled chains.

packages/indexer/src/database/BundleRepository.ts
```diff
@@ -168,7 +169,7 @@ export class BundleRepository extends utils.BaseRepository {
    * @param maxLookbackFromBlock The maximum number of blocks to look back from the provided block number (optional)
    * @returns The closest proposed (undisputed/non-canceled) root bundle back in time, or undefined if none are found
    */
-  public retrieveClosestProposedRootBundle(
+  public retrieveClosestProposedRootBundleEvent(
```
melisaguevara, line 172: +1 👍🏼

melisaguevara (commented):
Incredible work!
I left some comments

packages/indexer/src/database/BundleRepository.ts
james-a-morris: Done: bb1576a320becb8011122aea081e516c6377665e

packages/indexer/src/main.ts
james-a-morris: Done: 7c6d08eefd19d31b8cefebc5fe95cce879cccd8e

We have to keep the `bundleBuilderResult` prior to the spokeResults since `...spokeResults` is the remainder of the array

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -0,0 +1,235 @@
...
+        at: "BundleBuilder#Processor#initialize",
+        message: "Postgres connection not provided",
+      });
+      throw new Error("Postgres connection not provided");
+    }
+    this.bundleRepository = new BundleRepository(
+      this.config.postgres,
+      this.config.logger,
+      true,
+    );
```
amateima, line 59: Maybe we should replace the `postgres` constructor config with the repository. At least this was the pattern that was followed in the last implementations

packages/indexer/src/utils/bundleBuilderUtils.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+    throw new Error("No executed bundles found");
+  }
+  // If a proposed bundle is found, it must be newer than the last executed bundle
+  // otherwise, the system is in an inconsistent state
+  if (
+    !utils.isDefined(lastProposedBundle) ||
+    lastProposedBundle.proposal.blockNumber <
+      lastExecutedBundle.proposal.blockNumber
+  ) {
```
amateima, line 48: Shouldn't the condition be `utils.isDefined(lastProposedBundle) &&
    lastProposedBundle.proposal.blockNumber <
      lastExecutedBundle.proposal.blockNumber` ?

packages/indexer/src/utils/bundleBuilderUtils.ts
james-a-morris: Thoughts on something like the folowing? I like your suggestion but it will make the `startBlock === 1` in the event that `previousStartBlock === undefined`

```
/**
 * Given the previous and current proposed bundles, returns the block ranges for each chain.
 * @param previous The previous proposed bundle range.
 * @param current The current proposed bundle range.
 * @returns The block ranges for each chain. For each chain, this is the previous
 *          proposal's evaluation block number + 1 to the current proposal's evaluation block
 *          number. In the case that the new proposal includes ranges for a chain that was not
 *          previously included, the range starts at block 0 for that chain per the ACX UMIP. We
 *          also ensure that disabled chains aren't ticked up. I.e. if a chain has the same end
 *          block as the previous proposal range, we don't increment.
 * @dev The ordering of the ranges is the same as the chainIds in the current proposal.
 */
export function getBlockRangeBetweenBundles(
  previous: ProposalRange,
  current: ProposalRange
): ProposalRangeResult[] {
  return current.chainIds.map((chainId, idx) => ({
    chainId,
    startBlock: Math.min( // min to prevent ticking up disabled chains
      previous.bundleEvaluationBlockNumbers[idx]
        ? previous.bundleEvaluationBlockNumbers[idx] + 1
        : 0, // If this is a new chain, start from block 0
      current.bundleEvaluationBlockNumbers[idx]!
    ),
    endBlock: current.bundleEvaluationBlockNumbers[idx]!,
  }));
}
```

packages/indexer/src/services/bundles.ts
james-a-morris: Yep! This can all pop out 923fdceeb68d91f6d0cf3cbd6b8f65f7d9cc35c9 - good call - we should probably update the linter to catch unused vars

packages/indexer/src/utils/bundleBuilderUtils.ts
james-a-morris: Done: bf7bfbf0fcbbc27f4164ec4dcd3d7355fc7540a8

packages/indexer/src/services/bundleBuilder/index.ts
james-a-morris: We don't always have the roots available to us like in the call to `handleCurrentBundleLoop`

packages/indexer/src/utils/bundleBuilderUtils.ts
james-a-morris: This is tricky since we are ticking up with only the previous block range. One moment maybe we can add to this fn

packages/indexer/src/utils/bundleBuilderUtils.ts
```diff
@@ -0,0 +1,124 @@
...
+  previous: ProposalRange,
+  providers: RetryProvidersFactory,
+): Promise<ProposalRangeResult[]> {
+  return Promise.all(
+    previous.chainIds.map(async (chainId, idx) => {
+      const previousBlock = previous.bundleEvaluationBlockNumbers[idx];
+      const provider = providers.getProviderForChainId(chainId);
+      if (!utils.isDefined(provider)) {
+        throw new Error(`Provider for chain ${chainId} not found`);
+      }
```
amateima: I don't think this is needed as `getProviderForChainId` function throws if no provider is found

packages/indexer/src/utils/bundleBuilderUtils.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (!utils.isDefined(previousBlock)) {
+        throw new Error(`Previous block number for chain ${chainId} not found`);
+      }
+      const currentBlock = await provider.getBlockNumber();
+      return {
+        chainId,
+        startBlock: previousBlock + 1,
+        endBlock: currentBlock,
```
amateima: Should we check if the `previousBlock` is equal to `currentBlock`?

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -0,0 +1,235 @@
...
+      bundleHead,
+    );
+    // Instantiate spoke clients
+    const spokeClients = lookbackRange.reduce(
+      (acc, { chainId, startBlock, endBlock }) => ({
+        ...acc,
+        [chainId]: this.config.spokePoolClientFactory.get(
+          chainId,
+          startBlock,
+          endBlock,
```
amateima, line 175: Is this `endBlock` correct? Shouldn't it be end block of the range, not the one from the `historicalProposedBundle`?

packages/indexer/src/utils/bundleBuilderUtils.ts
james-a-morris: https://github.com/across-protocol/indexer/pull/47/commits/096290b2bbafcafd4dc2551387d70a5e3e7eb1f8

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -0,0 +1,235 @@
...
+  private bundleRepository: BundleRepository;
+
+  constructor(private config: BundleBuilderConfig) {
+    super(config.logger, "bundleBuilder");
+  }
+
+  protected async indexerLogic(): Promise<void> {
+    await Promise.allSettled([
+      this.handleCurrentBundleLoop(),
+      this.handleProposedBundleLoop(),
```
amateima, line 43: Instead of calling `resolveMostRecentProposedAndExecutedBundles` in each loop, can we call it a single time and pass the result as a parameter to the loops? This way we make sure that the loop functions are processing the same proposed and current bundles

packages/indexer/src/database/BundleRepository.ts
```diff
@@ -204,6 +205,37 @@ export class BundleRepository extends utils.BaseRepository {
...
+        blockNumber,
+      });
+    }
+    queryBuilder.orderBy("proposal.blockNumber", "DESC");
+    if (nthBundle !== undefined) {
+      queryBuilder.offset(nthBundle);
+    }
+    return queryBuilder.getOne();
+  }
+
```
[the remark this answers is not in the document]
james-a-morris: I'd like to keep the block numbers separate since this relation also includes disputes/cancelations that also have block numbers.

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -0,0 +1,235 @@
...
+  resolveMostRecentProposedAndExecutedBundles,
+  SpokePoolClientFactory,
+} from "../../utils";
+import { BundleRepository } from "../../database/BundleRepository";
+import { utils } from "@across-protocol/sdk";
+import Redis from "ioredis";
+import { RetryProvidersFactory } from "../../web3/RetryProvidersFactory";
+import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants";
+import { clients } from "@across-protocol/sdk";
+import { caching } from "@across-protocol/sdk";
```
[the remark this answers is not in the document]
james-a-morris: 6cd7f8eed2a4126890bbc5cd995597ed78bdb50f

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -0,0 +1,235 @@
...
+type BundleBuilderConfig = {
+  logger: winston.Logger;
+  postgres: DataSource;
+  redis: Redis;
+  providerFactory: RetryProvidersFactory;
+  hubClientFactory: HubPoolClientFactory;
+  spokePoolClientFactory: SpokePoolClientFactory;
+};
+
+export class Processor extends BaseIndexer {
```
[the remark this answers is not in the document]
james-a-morris: 0a83e3bbcc28f95383e252c90b452e35fc99b98c

packages/indexer/src/utils/bundleBuilderUtils.ts
james-a-morris: https://github.com/across-protocol/indexer/pull/47#discussion_r1803376983
https://github.com/across-protocol/indexer/pull/47#discussion_r1803343194

packages/indexer/src/services/bundleBuilder/index.ts
james-a-morris: I'm not sure what you mean. We create a historical lookback range from the end of the historical bundle (8 executed bundles prior) to the end of the block range in our bundle head lookback (will either be the head of the chain or the proposed bundle range end blocks)

```
    const historicalProposedBundle = historicalProposal.proposal;
    // Instantiate the Hub & ConfigStore Client from genesis
    const hubPoolClient = this.config.hubClientFactory.get(CHAIN_IDs.MAINNET);
    const configStoreClient = hubPoolClient.configStoreClient;
    // Resolve lookback range for the spoke clients
    const lookbackRange = getBlockRangeBetweenBundles(
      historicalProposedBundle,
      bundleHead,
    );
    // Instantiate spoke clients
    const spokeClients = lookbackRange.reduce(
      (acc, { chainId, startBlock, endBlock }) => ({
        ...acc,
        [chainId]: this.config.spokePoolClientFactory.get(
          chainId,
          startBlock,
          endBlock,
          {
            hubPoolClient,
          },
        ),
      }),
      {} as Record<number, clients.SpokePoolClient>,
    );
```

packages/indexer/src/utils/bundleBuilderUtils.ts
james-a-morris: Agreed - I had this stashed: https://github.com/across-protocol/indexer/pull/47/commits/180080e41bf264d79dd1dc306f67d641b7e1f3bc

packages/indexer/src/services/bundleBuilder/index.ts
melisaguevara: I see. Well let's keep it like this then

packages/indexer/src/services/bundleBuilder/index.ts
james-a-morris: https://github.com/across-protocol/indexer/pull/47/commits/4c438587998e8e7449e9ee0264d1d08e91144dc5

packages/indexer/src/services/bundleBuilder/index.ts
melisaguevara: When reconstructing older bundles I had to use slightly higher end blocks to instantiate the spokePoolClients. The reason is that it might happen that a fill within the bundle range corresponds to a deposit that's outside of the range for the destination chain.
For example, this [bundle](https://etherscan.io/tx/0x79fa46fbfd0134a1aba3d604c15564cbb0c818e9de7cfc9007fd87c5f6e91255#eventlog) includes this [fill](https://etherscan.io/tx/0xf2a036fba83fafba103b1fc2fd5a625b5141d92c5d9577826f2017dda14f46fc) that matches with this [deposit](https://lineascan.build/tx/0x700fbe7b20db55f2639f78cf3cb77a998ee8b9b36eb8ecf4a6cd18470358695b). But the deposit block number (10431813) is higher than the bundle end block for Linea (10431811).
So, if we update the spokeClients only up to the end block, such fills will be flagged as invalid.
I think this is not a problem when building the current bundle but it might be relevant when building the last proposed one.
The way I handled this was by adding a buffer equal to the average amount of blocks mined in 5 mins to the end block of each chain. I can share my solution with you if you think we should handle this now.

melisaguevara: synced offline. We'll modify this in an upcoming PR.

james-a-morris: cc @alexandrumatei36 we're planning on fast following this

melisaguevara (approved):

PR #49, opened 2024-09-25
feat(hubpool): set pool rebalance route for token mappings

james-a-morris:
adds tracking for `setrebalanceroute` events in the hubpool. This is used for dynamic tracking of l1Token:{l2Token|chainId} mappings


melisaguevara (commented):
Looks good! My only comment is that this event is named `SetPoolRebalanceRoute`.

james-a-morris:
> Looks good! My only comment is that this event is named `SetPoolRebalanceRoute`.

https://github.com/across-protocol/indexer/pull/49/commits/73484eaf6e854d07a18f5110697881460e30f18e

melisaguevara (approved):

packages/indexer-database/src/entities/evm/SetPoolRebalanceRoute.ts
```diff
@@ -0,0 +1,36 @@
...
+  l1Token: string;
+
+  @Column({ nullable: false })
+  destinationToken: string;
+
+  @Column({ nullable: false })
+  blockNumber: number;
+
+  @Column()
+  transactionHash: string;
```
amateima, line 26: should we have an UK constraint on these two columns?

amateima (approved):

PR #50, opened 2024-09-25
feat: implement generic Indexer with data finalisation

amateima:


packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -0,0 +1,79 @@
...
+export class Indexer {
+  constructor(
+    private config: ConstructorConfig,
+    private dataHandler: IndexerDataHandler,
+    private rpcProvider: ethers.providers.JsonRpcProvider,
+    private redisCache: RedisCache,
+    private logger: Logger,
+  ) {}
+
+  async start() {
```
daywiss, line 30: do you think we need some kind of stop function in order to safely shutdown the app?

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -0,0 +1,79 @@
...
+          lastBlockRangeFinalisedBlock,
+        );
+        await this.redisCache.set(
+          this.getLastFinalisedBlockCacheKey(),
+          lastBlockRangeFinalisedBlock,
+        );
+        await across.utils.delay(this.config.loopWaitTimeSeconds);
+      } catch (error) {
+        this.logger.error(`Indexer::Error processing block range`, error);
+        await across.utils.delay(this.config.loopWaitTimeSeconds);
```
daywiss, line 71: i wonder if this could be deduped in a `finally` block. dont remember if finally blocks async though

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  async start() {
+    while (true) {
+      try {
+        const lastBlockFinalised = await this.redisCache.get<number>(
+          this.getLastFinalisedBlockCacheKey(),
+        );
+        const fromBlock = lastBlockFinalised
+          ? lastBlockFinalised + 1
+          : this.dataHandler.getStartIndexingBlockNumber();
```
daywiss, line 38: wondering if this should be config for this class instead

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -0,0 +1,79 @@
...
+        await across.utils.delay(this.config.loopWaitTimeSeconds);
+      } catch (error) {
+        this.logger.error(`Indexer::Error processing block range`, error);
+        await across.utils.delay(this.config.loopWaitTimeSeconds);
+      }
+    }
+  }
+
+  private getLastFinalisedBlockCacheKey() {
+    return `indexer:lastBlockFinalised:${this.dataHandler.getDataIdentifier()}`;
```
daywiss, line 77: again it seems like this could just be passed into config for the class

daywiss (approved):
looks good, nice and simple, just a few interface comments

melisaguevara (approved):

packages/indexer/src/data-indexing/service/Indexer.ts
amateima: https://github.com/across-protocol/indexer/pull/50/commits/f8fd1a641830656e2b36e748db56e715854dbb86

amateima: also refactored the code a bit to make the start function cleaner

packages/indexer/src/data-indexing/service/Indexer.ts
amateima: True, but in my mind it made more sense to couple the data identifier with the logic used to fetch and process data, not with logic that manages the block ranges. This identifier will be in most cases a pair like `<contract_address>:<chain_id>`, so in case there's a need to change the contract address, only the `DataHandler` class will need to be changed and not touching the `Indexer` at all

packages/indexer/src/data-indexing/service/Indexer.ts
amateima: But I am not strongly opinionated on this. I will change it if it makes more sense for you to be part of the `Indexer` config 👍

packages/indexer/src/data-indexing/service/Indexer.ts
amateima: Missed this. Added https://github.com/across-protocol/indexer/pull/50/commits/4a7ae8b29367c45a58a886aaebd1edf8a46679e6

PR #51, opened 2024-09-30
fix: reduce retryprovider env variables

amateima:
- adds default values to many of the RetryProvider ENV variables
- implement a new `RetryProvidersFactory` class for instantiating the RetryProviders and using a single instance per chain


packages/indexer/src/parseEnv.ts
```diff
@@ -115,6 +117,68 @@ function parseRetryProviderConfig(
...
+      "0",
+  );
+  const providerCacheTtl = process.env.PROVIDER_CACHE_TTL
+    ? Number(process.env.PROVIDER_CACHE_TTL)
+    : undefined;
+  const nodeQuorumThreshold = Number(
+    process.env[`NODE_QUORUM_${chainId}`] || process.env.NODE_QUORUM || "1",
+  );
+  const retries = Number(
+    process.env[`NODE_RETRIES_${chainId}`] || process.env.NODE_RETRIES || "2",
```
daywiss, line 155: id probably set this to 1 or 0 (whatever no retries is) if you are omitting this

daywiss (approved):
looks good

packages/indexer/src/web3/RetryProvidersFactory.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  constructor(
+    private redisCache: RedisCache,
+    private logger: Logger,
+  ) {}
+
+  public initializeProviders() {
+    const providersUrls = parseProvidersUrls();
+
+    for (const chainId of this.retryProviders.keys()) {
```
melisaguevara, line 19: Should this use `providersUrls` keys instead?

melisaguevara (approved):

packages/indexer/src/web3/constants.ts
james-a-morris, line 1: Are these meant to be largely unchanged as time goes on?

james-a-morris (approved):
one question

packages/indexer/src/web3/RetryProvidersFactory.ts
amateima: Yes! Good catch! I actually fixed this in my upcoming PR but I forgot to fix it on the branch I was branching of 😄

packages/indexer/src/web3/constants.ts
amateima: Supposedly yes

PR #52, opened 2024-09-30
feat: migrate HubPoolIndexer to new Indexer

amateima:


packages/indexer-database/src/entities/evm/SetPoolRebalanceRoute.ts
```diff
@@ -7,7 +7,7 @@ import {
 } from "typeorm";
 
 @Entity({ schema: "evm" })
-@Unique("UK_setPoolRebalanceRoute_transactionHash_transactionIndex_logIndex", [
+@Unique("UK_spr_transactionHash_transactionIndex_logIndex", [
```
amateima, line 10: Constraints names have a max length of  63

packages/indexer/src/main.ts
```diff
@@ -54,14 +61,40 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
...
+      ),
+    },
+    hubPoolIndexerDataHandler,
+    retryProvidersFactory.getProviderForChainId(
+      acrossConstants.CHAIN_IDs.MAINNET,
+    ),
+    new RedisCache(redis),
+    logger,
+  );
+  await hubPoolIndexer.start();
```
amateima, line 88: later this will be moved inside an orchestrator class that will instantiate all indexers and indexerDataHandlers in order to keep the main file concise

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -0,0 +1,173 @@
...
+    });
+  }
+
+  private async fetchEventsByRange(
+    blockRange: BlockRange,
+  ): Promise<FetchEventsResult> {
+    const { hubPoolClient, configStoreClient } = this;
+
+    configStoreClient.eventSearchConfig.fromBlock = blockRange.from;
+    configStoreClient.eventSearchConfig.toBlock = blockRange.to;
```
melisaguevara, line 98: Is it ok to update the configStoreClient only for this block range? I remember @james-a-morris mentioned we should always update this client from deployment.

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  private async initialize() {
+    this.configStoreClient = await utils.getConfigStoreClient({
+      logger: this.logger,
+      provider: this.provider,
+      maxBlockLookBack: getMaxBlockLookBack(this.chainId),
+      chainId: this.chainId,
+    });
+    // how to limit the blocks to only the queried range?
```
melisaguevara, line 82: Is this something we still need to figure out?

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
amateima: Good catch 👍

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
amateima: removed

melisaguevara (approved):
lgtm!

PR #53, opened 2024-09-30
feat: add integrator id to deposit events

melisaguevara:
Add `integratorId` field to `v3FundsDeposited` events and `/deposits` endpoint query params.

### DIFF
--- a/packages/indexer-api/src/dtos/deposits.dto.ts
+++ b/packages/indexer-api/src/dtos/deposits.dto.ts
@@ -1,4 +1,5 @@
 import * as s from "superstruct";
+import { entities } from "@repo/indexer-database";
 
 const stringToInt = s.coerce(s.number(), s.string(), (value) =>
   parseInt(value),
@@ -7,14 +8,16 @@ const stringToInt = s.coerce(s.number(), s.string(), (value) =>
 export const DepositsParams = s.object({
   depositor: s.optional(s.string()),
   recipient: s.optional(s.string()),
+  originChainId: s.optional(stringToInt),
+  destinationChainId: s.optional(stringToInt),
   inputToken: s.optional(s.string()),
   outputToken: s.optional(s.string()),
-  integrator: s.optional(s.string()),
-  status: s.optional(s.string()),
+  integratorId: s.optional(s.string()),
+  status: s.optional(s.enums(Object.values(entities.RelayStatus))),
   // some kind of pagination options, skip could be the start point
   skip: s.optional(stringToInt),
   // pagination limit, how many to return after the start, note we convert string to number
-  limit: s.optional(stringToInt),
+  limit: s.defaulted(stringToInt, 50),
 });
 
 export type DepositsParams = s.Infer<typeof DepositsParams>;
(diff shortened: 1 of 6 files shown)


packages/indexer/src/services/spokePoolIndexer.ts
```diff
@@ -298,4 +308,21 @@ export class Indexer extends BaseIndexer {
       tokensBridgedEvents,
     };
   }
+
+  private async getIntegratorId(deposit: across.interfaces.DepositWithBlock) {
+    const INTEGRATOR_DELIMITER = "1dc0de";
```
daywiss, line 313: what does this come from?

packages/indexer-api/src/dtos/deposits.dto.ts
```diff
@@ -7,14 +8,16 @@ const stringToInt = s.coerce(s.number(), s.string(), (value) =>
 export const DepositsParams = s.object({
   depositor: s.optional(s.string()),
   recipient: s.optional(s.string()),
+  originChainId: s.optional(stringToInt),
+  destinationChainId: s.optional(stringToInt),
   inputToken: s.optional(s.string()),
   outputToken: s.optional(s.string()),
-  integrator: s.optional(s.string()),
-  status: s.optional(s.string()),
+  integratorId: s.optional(s.string()),
```
daywiss, line 15: what is an integrator?

packages/indexer/src/services/spokePoolIndexer.ts
```diff
@@ -298,4 +308,21 @@ export class Indexer extends BaseIndexer {
...
+    const INTEGRATOR_ID_LENGTH = 4; // Integrator ids are 4 characters long
+    let integratorId = undefined;
+    const txn = await this.spokePoolProvider.getTransaction(
+      deposit.transactionHash,
+    );
+    const txnData = txn.data;
+    if (txnData.includes(INTEGRATOR_DELIMITER)) {
```
daywiss, line 320: ok so this is some kind of referral program or something

daywiss (approved):

packages/indexer-api/src/dtos/deposits.dto.ts
amateima: @daywiss 
Across wants to have better insights about the flow of funds, so it was added an optional `integrator` flag that can be appended to the transaction call data and delimited by a hardcoded delimiter when calling SpokePool's `deposit()` function. [Here](https://docs.google.com/spreadsheets/d/1K4kyyYffjFn8LpTPu-p9ub0CL25iPcJ7c1cvvLJ1BI8/edit?gid=0#gid=0) we maintain a list of integrators IDs and the mapping to the platform it is appended by.
We store this in the DB as `/deposits` endpoint should be able to filter deposits by integrator

packages/indexer-api/src/dtos/deposits.dto.ts
```diff
@@ -7,14 +8,16 @@ const stringToInt = s.coerce(s.number(), s.string(), (value) =>
 export const DepositsParams = s.object({
   depositor: s.optional(s.string()),
   recipient: s.optional(s.string()),
+  originChainId: s.optional(stringToInt),
+  destinationChainId: s.optional(stringToInt),
```
amateima, line 12: why are these necessary when requesting lists of deposits?

amateima (approved):

PR #56, opened 2024-10-02
improve(factories): contract client factories

james-a-morris:
This update introduces factory classes that streamline the process of handling environment variable parsing and configuration storage for contract clients. By internalizing these operations, it becomes easier to create new clients without needing to pass around bulky config files, simplifying the setup and reducing redundancy.

### DIFF
--- a/packages/indexer/src/utils/contractFactoryUtils.ts
+++ b/packages/indexer/src/utils/contractFactoryUtils.ts
@@ -1,23 +1,18 @@
-import { Logger } from "winston";
-import { RetryProvidersFactory } from "../web3/RetryProvidersFactory";
+import { CHAIN_IDs } from "@across-protocol/constants";
 import { clients } from "@across-protocol/sdk";
+import { Logger } from "winston";
+import { getMaxBlockLookBack } from "../web3/constants";
+import { RetryProvidersFactory } from "../web3/RetryProvidersFactory";
 import {
   getConfigStoreClient,
   getHubPoolClient,
   getSpokeClient,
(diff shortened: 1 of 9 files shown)


packages/indexer/src/utils/contractFactoryUtils.ts
```diff
@@ -0,0 +1,149 @@
...
+  hubPoolClientFactory: HubPoolClientFactory;
+};
+type SpokeFactoryGetFunctionOverrides = {
+  hubPoolClient: clients.HubPoolClient;
+};
+export class SpokePoolClientFactory extends ContractClientFactory<
+  clients.SpokePoolClient,
+  SpokeFactoryRequiredFactories,
+  SpokeFactoryGetFunctionOverrides
+> {
```
melisaguevara, line 122: could this be missing here? or is it not necessary?
```
constructor(
    retryProviderFactory: RetryProvidersFactory,
    logger: Logger,
    requiredFactories: SpokeFactoryRequiredFactories,
  ) {
    super(retryProviderFactory, logger, requiredFactories);
  }
```

melisaguevara (approved):
This is really useful, thank you! 🙌🏼 
lgtm, just one question

packages/indexer/src/utils/contractFactoryUtils.ts
james-a-morris: I don't think it's exactly necessary but we should add it to ensure things are aligned

amateima (approved):

PR #63, opened 2024-10-04
fix(indexer): run hub indexer in parallel

james-a-morris:
This is currently set such that no other indexer runs.


daywiss (approved):

daywiss:
o is that cuz start is blocking now?

james-a-morris:
> o is that cuz start is blocking now?

yep - we have `start` in the `Promise.allSettled( )` so everything can gracefully end but this will blog infinitely

PR #64, opened 2024-10-07
feat: migrate SpokePoolIndexer to new Indexer

amateima:


packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -111,56 +130,100 @@ export class SpokePoolRepository extends utils.BaseRepository {
...
       };
     });
-    await this.insert(entities.TokensBridged, formattedEvents, throwError);
+    const chunkedEvents = across.utils.chunk(formattedEvents, this.chunkSize);
+    const savedEvents = await Promise.all(
+      chunkedEvents.map((eventsChunk) =>
+        this.insertWithFinalisationCheck(
+          entities.TokensBridged,
+          eventsChunk,
+          ["chainId", "leafId", "l2TokenAddress", "transactionHash"],
```
melisaguevara, line 222: If `transactionHash` changes due to the reorg, I'm thinking we might end having both the first event we saw and the finalised one in the database. But the delete unfinalised query should handle this case, right?

packages/indexer/src/services/SpokePoolIndexerDataHandler.ts
```diff
@@ -0,0 +1,235 @@
...
+    await this.spokePoolProcessor.process(storedEvents);
+  }
+
+  private async fetchEventsByRange(
+    blockRange: BlockRange,
+  ): Promise<FetchEventsResult> {
+    const { configStoreClient, hubPoolClient, spokePoolClient } = this;
+
+    configStoreClient.eventSearchConfig.toBlock = blockRange.to;
+    hubPoolClient.eventSearchConfig.toBlock = blockRange.to;
```
melisaguevara, line 104: this toBlock refers to the SpokePool's chain? Should we somehow translate that to mainnet block numbers?

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -67,33 +72,47 @@ export class SpokePoolRepository extends utils.BaseRepository {
...
       };
     });
     const chunkedEvents = across.utils.chunk(formattedEvents, this.chunkSize);
     const savedEvents = await Promise.all(
       chunkedEvents.map((eventsChunk) =>
-        this.insert(entities.FilledV3Relay, eventsChunk, throwError),
+        this.insertWithFinalisationCheck(
+          entities.FilledV3Relay,
+          eventsChunk,
+          ["depositId", "originChainId"],
```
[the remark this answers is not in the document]
amateima: 👍

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -111,56 +130,100 @@ export class SpokePoolRepository extends utils.BaseRepository {
...
-      entities.RequestedSpeedUpV3Deposit,
-      formattedEvents,
-      throwError,
+    const chunkedEvents = across.utils.chunk(formattedEvents, this.chunkSize);
+    const savedEvents = await Promise.all(
+      chunkedEvents.map((eventsChunk) =>
+        this.insertWithFinalisationCheck(
+          entities.RequestedSpeedUpV3Deposit,
+          eventsChunk,
+          ["depositId", "originChainId"],
```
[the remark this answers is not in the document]
amateima: I added the transaction hash as the 3rd component of the unique key 👍

packages/indexer/src/database/SpokePoolRepository.ts
amateima: > But the delete unfinalised query should handle this case, right?
Good observation. and yes, the delete unfinalised query should handle it

packages/indexer/src/services/SpokePoolIndexerDataHandler.ts
amateima: Ah, you're right. I removed the to blocks for the config client and hub client

melisaguevara (approved):
lgtm! 😄

PR #69, opened 2024-10-14
fix(bundles): fix range start block and bundle id when assigning events

melisaguevara:
This PR fixes two issues related to bundles:
- When saving the bundle range, we were not incrementing the start block so the end block of each bundle was being repeated as the start block of the next bundle.
- When relating events to bundles, we were assuming that the bundle id was equal to the id of the proposal, but that might not be true in some cases.

### DIFF
--- a/packages/indexer-database/src/entities/evm/RootBundleCanceled.ts
+++ b/packages/indexer-database/src/entities/evm/RootBundleCanceled.ts
@@ -2,9 +2,11 @@ import {
   Column,
   CreateDateColumn,
   Entity,
+  OneToOne,
   PrimaryGeneratedColumn,
   Unique,
 } from "typeorm";
+import { Bundle } from "../Bundle";
 
 @Entity({ schema: "evm" })
 @Unique("UK_rootBundleCanceled_txHash", ["transactionHash"])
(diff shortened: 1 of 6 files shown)


james-a-morris:
OOC which cases would the ids not match? These are foreign keys to incremental ids

packages/indexer-database/src/entities/evm/RootBundleCanceled.ts
```diff
@@ -18,6 +20,9 @@ export class RootBundleCanceled {
   @Column()
   requestTime: Date;
 
+  @OneToOne(() => Bundle, (bundle) => bundle.cancelation)
```
james-a-morris, line 23: We would need this to be nullable. I thought we didn't want to change these raw event entities?

packages/indexer/src/database/BundleRepository.ts
```diff
@@ -48,14 +48,14 @@ export class BundleRepository extends utils.BaseRepository {
...
+      .createQueryBuilder("crb")
       .select([
-        "drb.id",
-        "drb.blockNumber",
-        "drb.logIndex",
-        "drb.transactionIndex",
+        "crb.id",
+        "crb.blockNumber",
+        "crb.logIndex",
+        "crb.transactionIndex",
```
james-a-morris, line 56: Fair call but is out of scope

james-a-morris (approved):

PR #71, opened 2024-10-15
fix: remove the logic for deleting removed events

amateima:
This query was meant to delete events that were removed due to a re-org, but the query is wrong and it deletes more data than it should


melisaguevara (approved):
Just OOC, do you know why it was deleting more data than expected?

amateima:
> Just OOC, do you know why it was deleting more data than expected?

Because deleting by block number only is a mistake. You also need to filter by other parameters, such as `chainId` because each chain has its own last finalised block (think about `deposit` events for example)

PR #74, opened 2024-10-15
fix: optimise spoke pool client usage

amateima:
This PR removes the spokeClient instance that lives inside the SpokePoolIndexerDataHandler class, so it doesn't accumulate events anymore other than just those needed from the fetched block range


melisaguevara (approved):
nice catch!

PR #75, opened 2024-10-15
feat: add indexed bundle leaf cache in redis

daywiss:
# motivation
We need to cache bundle leaf data in a flexible way to query from the api

# changes
This adds a redis class to cache the basic data as well as allow queries by chain id, token or chain id and token


packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -79,6 +90,20 @@ export class Processor extends BaseIndexer {
       convertProposalRangeResultToProposalRange(ranges),
     );
     // Persist this to Redis
+    await Promise.all(
+      resultsToPersist
```
james-a-morris, line 94: Could we do this as a flat map?

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -79,6 +90,20 @@ export class Processor extends BaseIndexer {
...
+        .map((leaf) => {
+          return leaf.l1Tokens.map((l1Token, tokenIndex) => {
+            return this.currentBundleCache.set({
+              chainId: leaf.chainId,
+              l1Token,
+              netSendAmount: leaf.netSendAmounts[tokenIndex] ?? "0",
```
james-a-morris, line 100: We should probably throw in a case like this

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -110,6 +135,20 @@ export class Processor extends BaseIndexer {
...
     // Persist this to Redis
+    await Promise.all(
+      resultsToPersist
+        .map((leaf) => {
+          return leaf.l1Tokens.map((l1Token, tokenIndex) => {
+            return this.proposedBundleCache.set({
+              chainId: leaf.chainId,
+              l1Token,
+              netSendAmount: leaf.netSendAmounts[tokenIndex] ?? "0",
+              runningBalance: leaf.runningBalances[tokenIndex] ?? "0",
```
james-a-morris, line 146: Similar to above

packages/indexer/src/redis/bundleLeavesCache.ts
```diff
@@ -0,0 +1,124 @@
...
+  // Delete a record by chainId and l1Token
+  async delete(chainId: number, l1Token: string): Promise<boolean> {
+    const key = this.getKey(chainId, l1Token);
+
+    // Remove from Redis
+    const result = await this.config.redis.del(key);
+
+    // Also remove from the indexes
+    await this.config.redis.srem(this.getChainIdIndexKey(chainId), key);
+    await this.config.redis.srem(this.getL1TokenIndexKey(l1Token), key);
```
james-a-morris, line 62: Does this mean we're tracking two keys? Could we add this in a comment at the top or maybe with JSDoc comments on each function?

packages/indexer/src/services/bundleBuilder/index.ts
daywiss: this is already using map and flat, not sure what you mean

packages/indexer/src/redis/bundleLeavesCache.ts
daywiss: added docs

packages/indexer/src/services/bundleBuilder/index.ts
daywiss: added checks for this

melisaguevara (approved):

packages/indexer/src/services/bundleBuilder/index.ts
james-a-morris: I.e. `flatMap( () => {})` vs `map().flat()`

packages/indexer/src/services/bundleBuilder/index.ts
```diff
@@ -110,6 +144,28 @@ export class Processor extends BaseIndexer {
...
+          return leaf.l1Tokens.map((l1Token, tokenIndex) => {
+            return this.proposedBundleCache.set({
+              chainId: leaf.chainId,
+              l1Token,
+              netSendAmount: leaf.netSendAmounts[tokenIndex]!,
+              runningBalance: leaf.runningBalances[tokenIndex]!,
+            });
+          });
+        })
+        .flat(),
```
james-a-morris, line 167: I'm not sure we'd want to store the l1token/chainId individually. The API will either return all l1Tokens for a given chainId or a filtration of just 1 l1Token if a query parameter is set.

We'd want to store the entire set of 3 arrays in one element otherwise we'll need to query `proposedBundleCache:1:*` for each `chainId`.

packages/indexer/src/redis/bundleLeavesCache.ts
```diff
@@ -0,0 +1,196 @@
...
+   * @param {BundleLeaf} data - The BundleLeaf data to store.
+   * @returns {Promise<void>} - A promise that resolves when the data is successfully stored.
+   */
+  async set(data: BundleLeaf): Promise<void> {
+    const key = this.getKey(data.chainId, data.l1Token);
+    await this.config.redis.set(key, JSON.stringify(data));
+
+    // Add to indexes for quick retrieval by chainId and l1Token separately
+    await this.config.redis.sadd(this.getChainIdIndexKey(data.chainId), key);
+    await this.config.redis.sadd(this.getL1TokenIndexKey(data.l1Token), key);
```
james-a-morris, line 41: Not all `l1Tokens` are enabled for all `chainIds`

daywiss: good call, this probably means we should clear cache before storing next update

packages/indexer/src/services/bundleBuilder/index.ts
daywiss: the idea here was you could efficiently query the list and do a reduce operation in the api since we are indexing by chain/token and chain+token.

"API will either return all l1Tokens for a given chainId" - for this query you can call `cache.getByChainId(chainId)` which returns results for all tokens for a single chain
"a filtration of just 1 l1Token" - for this you query `cache.getByL1Token(tokenid)` which returns you all results for a single token across all chains

the api makes theese queries and needs to reduce the list to whatever form it needed to return to client

packages/indexer/src/redis/bundleLeavesCache.ts
daywiss: updated this to clear cache before next update

amateima (approved):
Solid work @daywiss ! fyi, the bundleBuilder index file has been renamed to `BundleBuilderService`

james-a-morris (approved):

PR #78, opened 2024-10-17
feat: hubpool balance tracking and persistence

james-a-morris:
Adds aggregated hub balance persistence to bundle builder service for consumption in `/hubbalance` api endpoint

PR #81, opened 2024-10-22
feat(indexer-api): expose hubpool balance endpoint

daywiss:
# Motivation
Expose hubpool cached balance values through the indexer api

# Changes
This basically just forwards data stored in `hubBalancesCache` from the indexer and can be accessed at `GET hubPoolBalances`

This is dependent on #78 getting in first


melisaguevara (approved):
LGTM

packages/indexer-api/src/dtos/balances.dto.ts
```diff
@@ -2,8 +2,7 @@ import * as s from "superstruct";
 
 // query hub pools by chainId? default to 1 if not specified. will leave option in case of testnets?
 export const HubPoolBalanceParams = s.object({
```
amateima, line 4: Can we name this `HubPoolBalanceParams` to make clear that these are query params and not URL path components?

packages/indexer-api/src/routers/balances.ts
```diff
@@ -0,0 +1,13 @@
...
+import Redis from "ioredis";
+import { BalancesController } from "../controllers/balances";
+import { BalancesService } from "../services/balances";
+
+export function getRouter(redis: Redis): Router {
+  const router = Router();
+  const service = new BalancesService(redis);
+  const controller = new BalancesController(service);
+  router.get("/hubPoolBalance", controller.getHubPoolBalance);
+  router.get("/spokePoolBalance", controller.getSpokePoolBalance);
```
amateima, line 11: I think the REST good practice is to not use camel case in the urls
```suggestion
  router.get("/hubpool-balance", controller.getHubPoolBalance);
  router.get("/spokepool-balance", controller.getSpokePoolBalance);
```

amateima (approved):

packages/indexer-api/src/dtos/balances.dto.ts
daywiss: i dont understand what you mean

PR #82, opened 2024-10-22
feat(api balance endpoint): add spoke balance API endpoint

james-a-morris:
Dependent on @daywiss's #81 endpoint. Adds spoke balance tracking.

### DIFF
--- a/packages/indexer-api/src/dtos/balances.dto.ts
+++ b/packages/indexer-api/src/dtos/balances.dto.ts
@@ -10,7 +10,7 @@ export const SpokePoolBalanceParams = s.object({
   chainId: s.number(),
   l1Token: s.optional(s.string()),
   // unsure why we have timestamp, implies we are storign history of balances? this is in the spec.
-  timestamp: s.number(),
+  timestamp: s.optional(s.number()),
 });
 
 export type SpokePoolBalanceResultElement = {
--- a/packages/indexer-api/src/dtos/balances.dto.ts
+++ b/packages/indexer-api/src/dtos/balances.dto.ts
@@ -6,7 +6,7 @@ export const HubPoolBalanceQueryParams = s.object({
 });
 
 // query spokepools by chainId, must specify
-export const SpokePoolBalanceParams = s.object({
+export const SpokePoolBalanceQueryParams = s.object({
   chainId: s.number(),
   l1Token: s.optional(s.string()),
 });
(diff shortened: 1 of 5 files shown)


packages/indexer-api/src/dtos/balances.dto.ts
```diff
@@ -2,15 +2,25 @@ import * as s from "superstruct";
...
-  l1Token: s.string(),
+  l1Token: s.optional(s.string()),
 });
 
 // query spokepools by chainId, must specify
 export const SpokePoolBalanceParams = s.object({
   chainId: s.number(),
+  l1Token: s.optional(s.string()),
   // unsure why we have timestamp, implies we are storign history of balances? this is in the spec.
   timestamp: s.number(),
```
melisaguevara, line 13: Seems like we are not using this, should we delete it?

melisaguevara (commented):
LGTM! Just one question

packages/indexer-api/src/dtos/balances.dto.ts
james-a-morris: It's in the spec so I'm inclined to keep it. cc @amateima

packages/indexer-api/src/dtos/balances.dto.ts
```diff
@@ -8,8 +8,19 @@ export const HubPoolBalanceQueryParams = s.object({
 // query spokepools by chainId, must specify
 export const SpokePoolBalanceParams = s.object({
   chainId: s.number(),
+  l1Token: s.optional(s.string()),
   // unsure why we have timestamp, implies we are storign history of balances? this is in the spec.
-  timestamp: s.number(),
-  // unsure why specified as l2Token in spec, don't we have spoke pool on L1?
-  l2Token: s.optional(s.number()),
+  timestamp: s.optional(s.number()),
```
amateima, line 13: we can remove this for now

amateima (approved):

packages/indexer-api/src/dtos/balances.dto.ts
```diff
@@ -8,8 +8,19 @@ export const HubPoolBalanceQueryParams = s.object({
 // query spokepools by chainId, must specify
 export const SpokePoolBalanceParams = s.object({
```
daywiss, line 9: one nit here was to rename this to `SpokePoolBalanceQueryParams`

daywiss (approved):
pretty cool

amateima (approved):

packages/indexer-api/src/dtos/balances.dto.ts
james-a-morris: Done

PR #83, opened 2024-10-23
feat: fetch integrator id asynchronously when backfilling

melisaguevara:


packages/indexer/src/messaging/IntegratorIdWorker.ts
```diff
@@ -0,0 +1,64 @@
...
+    private postgres: DataSource,
+    private logger: winston.Logger,
+    private providerFactory: RetryProvidersFactory,
+  ) {
+    this.setWorker();
+  }
+
+  public setWorker() {
+    this.worker = new Worker(
+      IndexerQueues.IntegratorId,
```
daywiss, line 29: could you put a comment on what this is doing, from what i can see its tryign to attach and integrator id to a deposit?

packages/indexer/src/services/SpokePoolIndexerDataHandler.ts
```diff
@@ -122,14 +125,25 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
...
-      v3FundsDepositedEvents.map(async (deposit) => {
-        return {
-          ...deposit,
-          integratorId: await this.getIntegratorId(deposit),
-        };
-      }),
-    );
+    let v3FundsDepositedWithIntegradorId;
+    // If fewer than 1000 deposits, fetch integrator ID synchronously.
+    // For larger sets, use the IntegratorId queue for asynchronous processing.
```
daywiss, line 130: why not just do them all async? keeps it a bit simpler

daywiss (approved):

packages/indexer/src/messaging/IntegratorIdWorker.ts
melisaguevara: Exactly! I added some docs here [9e1998f](https://github.com/across-protocol/indexer/pull/83/commits/9e1998f7707307ee507c57e32f6181dd4a823b48)

packages/indexer/src/services/SpokePoolIndexerDataHandler.ts
melisaguevara: Mainly because we want to have the integratorId available as soon as we store deposits.
The idea is just to reduce the load from this calls for tx data when backfilling.

melisaguevara:
@amateima Based on our discussion:
- The logic for populating the integrator id is now placed on the `processBlockRange` function and the messages are published to the queue after the deposits are stored.
- The integratorId message is now using only the relayHash.
- If the deposit is not found in the database or if it already has an integratorId associated we return.

amateima (approved):

PR #84, opened 2024-10-23
fix: slow fills insertion

melisaguevara:
- Deleting `fromLiteChain` and `toLiteChain` columns from `requestedV3SlowFill` table as they are not populated for slow fills by the SpokePool clients.
- Changing the unique keys we are using when inserting slow fills with finalisation check as they have to match the unique constraint of the table.

### DIFF
--- a/packages/indexer/src/database/SpokePoolRepository.ts
+++ b/packages/indexer/src/database/SpokePoolRepository.ts
@@ -115,7 +115,7 @@ export class SpokePoolRepository extends utils.BaseRepository {
         this.insertWithFinalisationCheck(
           entities.RequestedV3SlowFill,
           eventsChunk,
-          ["depositId", "originChainId"],
+          ["relayHash"],
           lastFinalisedBlock,
         ),
       ),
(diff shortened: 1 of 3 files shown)


amateima:
`fromLiteChain` and `toLiteChain` are populated only for deposit events, right @melisaguevara ?

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -115,7 +115,7 @@ export class SpokePoolRepository extends utils.BaseRepository {
         this.insertWithFinalisationCheck(
           entities.RequestedV3SlowFill,
           eventsChunk,
-          ["depositId", "originChainId"],
+          ["relayHash"],
```
amateima, line 118: Can a `relayHash`have multiple speedup requests?

melisaguevara:
> `fromLiteChain` and `toLiteChain` are populated only for deposit events, right @melisaguevara ?

Yes exactly.

packages/indexer/src/database/SpokePoolRepository.ts
melisaguevara: This one is for slow fill requests but yes, there can be only one slowFill request per relayHash. Here: https://github.com/across-protocol/contracts/blob/master/contracts/SpokePool.sol#L938

amateima (approved):

daywiss (approved):

PR #89, opened 2024-10-29
fix: remove duplicated events and add logs

amateima:


packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -76,7 +74,15 @@ export class HubPoolIndexerDataHandler implements IndexerDataHandler {
       blockRange,
       identifier: this.getDataIdentifier(),
     });
-    await this.storeEvents(events, lastFinalisedBlock);
+    const eventsWithoutDuplicates = await this.removeDuplicateEvents(events);
```
daywiss, line 77: how would we get duplicate events??

daywiss (approved):
nice find

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
melisaguevara: i'm also curious about that

amateima: I think the problem is deeper that I initially thought. If you request events for the same block range multiple times, the SDK clients will actually append the duplicated events to the ones already stored. This doesn't play well with our logic that does retries in case of errors

amateima:
I removed the logic added in previous commits and kept only the new logs

PR #93, opened 2024-10-31
fix: avoid hub client update multiple times for the same block range

amateima:


packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -32,35 +38,43 @@ export class Indexer {
   }
 
   public async start() {
+    let blockRangeResult: BlockRangeResult | undefined = undefined;
+    let blockRangeProcessedSuccessfully = true;
```
amateima, line 42: we store the result of the getBlockRange() function outside of the while loop in order to have access to previous block ranges. This will help us to not move forward to the block range if the previous block range was not processed successfully

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -32,35 +38,43 @@ export class Indexer {
...
         }
+        blockRangeProcessedSuccessfully = true;
       } catch (error) {
         this.logger.error({
           at: "Indexer::start",
           message: "Error processing block range",
           dataIdentifier: this.dataHandler.getDataIdentifier(),
           error,
         });
+        blockRangeProcessedSuccessfully = false;
```
amateima, line 77: If an error happened along the way in processing the block range, then don't compute a new `to` block, but process the exact previous block range again

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -60,7 +61,19 @@ export class HubPoolIndexerDataHandler implements IndexerDataHandler {
...
+
+    if (this.cachedFetchEventsResult) {
+      this.logger.info({
+        at: "HubPoolIndexerDataHandler::processBlockRange",
+        message: `Using cached events for ${this.getDataIdentifier()}`,
+      });
+      events = this.cachedFetchEventsResult;
+    } else {
+      events = await this.fetchEventsByRange(blockRange);
+      this.cachedFetchEventsResult = events;
```
amateima, line 74: if we fetch the events but errors are thrown after, cache the events so that we don't update the `hubPoolClient` for the same block range. This would cause to have duplicated events stored in the hub client

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -82,6 +95,7 @@ export class HubPoolIndexerDataHandler implements IndexerDataHandler {
       lastFinalisedBlock,
       identifier: this.getDataIdentifier(),
     });
+    this.cachedFetchEventsResult = undefined;
```
amateima, line 98: Once events are processed, delete them from cache

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -101,7 +115,6 @@ export class HubPoolIndexerDataHandler implements IndexerDataHandler {
   ): Promise<FetchEventsResult> {
     const { hubPoolClient, configStoreClient } = this;
 
-    hubPoolClient.firstBlockToSearch = blockRange.from;
```
amateima, line 104: we actually need all hub pool events from deployment block

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -32,35 +38,43 @@ export class Indexer {
...
-        const { latestBlockNumber, blockRange, lastFinalisedBlock } =
-          await this.getBlockRange();
+        // if the previous block range was processed successfully or if this is the first loop iteration,
+        // get the next block range to process
+        if (blockRangeProcessedSuccessfully || !blockRangeResult) {
+          blockRangeResult = await this.getBlockRange();
+        }
 
-        if (!blockRange) {
+        if (!blockRangeResult?.blockRange) {
```
james-a-morris, line 52: Would it make sense to `blockRangeProcessedSuccessfully = false` here to confirm that the block range wasn't processed

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -60,7 +61,19 @@ export class HubPoolIndexerDataHandler implements IndexerDataHandler {
       await this.initialize();
       this.isInitialized = true;
     }
-    const events = await this.fetchEventsByRange(blockRange);
+    let events: FetchEventsResult;
+
+    if (this.cachedFetchEventsResult) {
+      this.logger.info({
```
james-a-morris, line 67: Should this be more of a debug or does it make sense to log to GCP

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -60,7 +61,19 @@ export class HubPoolIndexerDataHandler implements IndexerDataHandler {
       await this.initialize();
       this.isInitialized = true;
     }
-    const events = await this.fetchEventsByRange(blockRange);
+    let events: FetchEventsResult;
```
james-a-morris, line 64: Should we store this in redis to make it persistent across runs?

james-a-morris (commented):
few comments + nit

packages/indexer/src/data-indexing/service/Indexer.ts
amateima: I don't think so because by setting `blockRangeProcessedSuccessfully` to `false` we would prevent `getBlockRange()` to be called again and we want that to happen because `blockRangeResult.blockRange` is undefined only when no new block was mined, but we continue to call `getBlockRange()` until new blocks are mined

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
amateima: In my mind it makes sense to log it in GCP

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
amateima: what to you mean by persistent across runs? You mean when the indexer is redeployed?

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -32,35 +38,43 @@ export class Indexer {
   }
 
   public async start() {
+    let blockRangeResult: BlockRangeResult | undefined = undefined;
```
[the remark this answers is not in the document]
amateima: Screenshot 2024-10-31 at 22 47 50


this is not valid class property declaration

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
amateima: Also, what I do in GCP is to filter logs to have severity > debug, so that I don't get all the SDK logs

packages/indexer/src/data-indexing/service/Indexer.ts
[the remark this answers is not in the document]
james-a-morris: Oh wow didn't realize

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
james-a-morris: Correct

james-a-morris (commented):
Replied above

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
amateima: there's no need to do that because if the indexer is redeployed then the HubPool client instance will start from clean state, so no need to worry about events duplication

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -32,35 +38,43 @@ export class Indexer {
...
+    let blockRangeProcessedSuccessfully = true;
+
     while (!this.stopRequested) {
       try {
-        const { latestBlockNumber, blockRange, lastFinalisedBlock } =
-          await this.getBlockRange();
+        // if the previous block range was processed successfully or if this is the first loop iteration,
+        // get the next block range to process
+        if (blockRangeProcessedSuccessfully || !blockRangeResult) {
+          blockRangeResult = await this.getBlockRange();
```
daywiss, line 49: i guess if we dont enter this condition, its retrying the previous range?

daywiss (approved):
hard to reason about how this will work, but if it looks good from testing, ok

packages/indexer/src/data-indexing/service/Indexer.ts
amateima: yes

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
```diff
@@ -60,7 +61,19 @@ export class HubPoolIndexerDataHandler implements IndexerDataHandler {
...
+
+    if (this.cachedFetchEventsResult) {
```
melisaguevara, line 66: Should we use this cache logic also for the `SpokePoolIndexerDataHandler`?

packages/indexer/src/data-indexing/service/Indexer.ts
melisaguevara: Would we need to get alerted somehow when this happens? To avoid long delays if something's wrong

packages/indexer/src/services/HubPoolIndexerDataHandler.ts
amateima: > Should we use this cache logic also for the SpokePoolIndexerDataHandler? Or is this a problem only in the SpokePoolClient?
I don't think it's necessary because the `SpokePoolIndexerDataHandler` re-instantiate the `spokeClient` instance on each run

PR #95, opened 2024-11-06
feat(benchmark): add very basic benchmarking

daywiss:
# motivation
we want to know how long it takes to process deposit events

# changes
1. This adds a couple things, a benchmarking class where we can start/stop unique ids
2. an independent contract listener, listening to deposit events, to kick off benchmark start
3. callbacks to spoke pool processor that fires once all events are processed, which kick off benchmark stop
4. wires this together in indexer/main and prints this on an interval to logger


james-a-morris (commented):
OOC do [these changes](https://github.com/across-protocol/relayer/pull/1896) intersect with the code we're adding here? If so - should we colocate it in the SDK?

daywiss:
> OOC do [these changes](https://github.com/across-protocol/relayer/pull/1896) intersect with the code we're adding here? If so - should we colocate it in the SDK?

looks like theres overlap. didnt know anything about this, will leave up to alex to decide how to proceed with this

packages/indexer/src/main.ts
```diff
@@ -95,7 +113,32 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
     new SpokePoolRepository(postgres, logger),
     redisCache,
     indexerQueuesService,
-    write,
+    (params: { type: WebhookTypes; event: JSONValue }) => {
```
daywiss, line 116: we are repurposing the webhook callback to also stop any deposit benchmarks

packages/indexer/src/main.ts
```diff
@@ -117,6 +160,11 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
     retryProvidersFactory,
   );
 
+  const stopDepositListener = listenForDeposits(
```
daywiss, line 163: this is a separate listener that listens straight on the contract for new deposits, given providers on various chains

packages/indexer/src/utils/benchmarks.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function getSpokepoolContract(params: GetSpokeClientParams) {
+  return SpokePoolFactory.connect(params.address, params.provider);
+}
+export type ProviderChainId = {
+  provider: providers.Provider;
+  chainId: number;
+};
+
+export function listenForDeposits(
```
daywiss, line 26: this function takes providers and listens to specific events on the spoke pool. it does this to start profiling a deposit. it will end when indexer finds and updates it

packages/indexer/src/main.ts
```diff
@@ -95,7 +113,32 @@ export async function Main(config: parseEnv.Config, logger: winston.Logger) {
...
-    write,
+    (params: { type: WebhookTypes; event: JSONValue }) => {
+      // stop any benchmarks based on origin and deposit it
+      if (params.type === WebhookTypes.DepositStatus) {
+        const depositStatusEvent =
+          params.event as eventProcessors.DepositStatusEvent;
+        const uniqueId = `${depositStatusEvent.originChainId}-${depositStatusEvent.depositId}`;
+        try {
+          const duration = depositBenchmark.end(uniqueId);
+          logger.debug({
```
daywiss, line 124: heres where we log the time. duration is in MS since start was called

melisaguevara (approved):

daywiss:
closing this in favor of #129

PR #98, opened 2024-11-11
fix: fix bundle events processor

amateima:


packages/indexer/src/database/BundleRepository.ts
```diff
@@ -171,40 +176,69 @@ export class BundleRepository extends utils.BaseRepository {
...
-          // If maxLookbackFromBlock is undefined, then allow the full range of blocks to be searched
-          blockDiff:
-            maxLookbackFromBlock !== undefined
-              ? blockNumber - maxLookbackFromBlock
-              : 0,
-        },
-      )
-      .orderBy("prb.blockNumber", "DESC") // Grab the most recent proposal
-      .getOne();
+      .findOne({
```
amateima, line 187: Replace createQueryBuilder() with findOne() because of a bug that prevents queries to be executed in parallel

pxrl (approved):

PR #99, opened 2024-11-12
fix: save bundle included events using a separate class

amateima:
Put the logic for saving bundle events in a new class `BundleIncludedEventsService`


melisaguevara (approved):
lgtm

PR #100, opened 2024-11-12
feat: add stats page

amateima:
Quick PR to add a stats page `/pages/stats`

Screenshot 2024-11-12 at 13 27 12


packages/indexer-api/src/services/stats-page.ts
```diff
@@ -0,0 +1,14 @@
+import { DataSource, entities } from "@repo/indexer-database";
+
+export class StatsPageService {
+  constructor(private db: DataSource) {}
+
+  public async getStatsPageData() {
+    const bundles = await this.db.getRepository(entities.Bundle).find({
+      relations: { proposal: true },
+      order: { proposal: { blockNumber: "DESC" } },
+      take: 6,
```
daywiss, line 10: coudl this be turned into query params?

daywiss (approved):
cool idea

packages/indexer-api/src/services/stats-page.ts
amateima: eventually yes

PR #101, opened 2024-11-14
feat: webhook module v1

daywiss:
# Motivation
We want to be able to handle webhook requests to the indexer, starting with deposit status

# changes
This adds a module in the indexer to startup an express based webhook manager, with an interface to add new event types, as well as register many webhooks. 

# usage
`factory` in the main entry point which gives you a  `WebhookFactory` call, which returns `{webooks ,express ,notifier}`.  As the indexer, all you need to do is call webhooks.write({type:'DepositStatus',event:EventData})
You also need to setup an interval to run the notification worker by calling `notifier.tick()` which will try to send out all pending notifications to webhooks

integrators will call into the express app on `POST /webhook`, passing body = `{type:'DepositStatus',filter:{originChainId,depositTxHash}, url:string}`. This will begin the listening process for indexer updates. It should also emit a first deposit status if one exists.

# testing
this has not been tested

# TODO
In future prs:
1. some testing
2. better api gating and client management
3. add persistence to stores


packages/indexer/src/webhooks/notifier.ts
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  public tick = async (now = Date.now()): Promise<void> => {
+    for await (const notification of this.pending.values()) {
+      if (notification.nextTry <= now) {
+        try {
+          await this.deps.notify(
+            notification.payload.url,
+            notification.payload.event,
+          );
```
daywiss, line 62: this is where the webhook callback happens. further down you can see we try a backoff strategy if it fails

packages/indexer/src/webhooks/webhook.ts
```diff
@@ -0,0 +1,91 @@
...
+export type DepositStatusNotification = {
+  hook: Webhook;
+  event: DepositStatusEvent;
+};
+export type Dependencies = {
+  hooks: WebhookRequests;
+  notify: (params: NotificationPayload) => void;
+  postgres: DataSource;
+};
+export class DepositStatusWebhook implements IWebhook {
```
daywiss, line 30: this is the specific deposit status processor we care about, its wrapped in a higher level webhooks class in case we want to define different types

packages/indexer/src/webhooks/factory.ts
```diff
@@ -0,0 +1,59 @@
...
+    requireApiKey: boolean;
+  };
+  enable: string[];
+};
+type Dependencies = {
+  postgres: DataSource;
+  logger: Logger;
+};
+
+export function WebhookFactory(config: Config, deps: Dependencies) {
```
daywiss, line 25: main entry point as an indexer

packages/indexer/src/webhooks/factory.ts
```diff
@@ -0,0 +1,59 @@
...
+      // add more webhook types here
+      case "DepositStatus": {
+        webhooks.registerWebhookProcessor(
+          name,
+          new DepositStatusWebhook({
+            postgres,
+            hooks,
+            notify: notifier.notify,
+          }),
+        );
```
daywiss, line 49: this is where we instance and register each webhook event type, currently we only have 'DepositStatus'
we can see it has a callback `notify` which queues it in the notifier, which persists notification state

packages/indexer/src/webhooks/notifier.ts
```diff
@@ -0,0 +1,102 @@
...
+    await this.pending.set(id, notification);
+    return id;
+  };
+  public notify = (payload: NotificationPayload): void => {
+    this.create(payload).catch((error) => {
+      this.logger.error(`Error creating notification:`, error);
+    });
+  };
+
+  public tick = async (now = Date.now()): Promise<void> => {
```
daywiss, line 55: you will manually need to call notifier.tick on an interval somewhere in indexer code ( or whoever consumes this module)

packages/indexer/src/webhooks/notifier.ts
```diff
@@ -0,0 +1,102 @@
...
+    for await (const notification of this.completed.values()) {
+      completedNotifications.push(notification);
+    }
+    return completedNotifications;
+  }
+}
+
+export class WebhookNotifier extends BaseNotifier {
+  constructor(deps: Omit<Dependencies, "notify">) {
+    super({ ...deps, notify: post });
```
daywiss, line 100: makes a specific webhook caller, we could theoretically pass in an event emitter for testing rather than https post

packages/indexer/src/webhooks/store.ts
```diff
@@ -0,0 +1,107 @@
+import { createClient, RedisClientType } from "redis";
+
+export interface AsyncStore<V> {
```
daywiss, line 3: abstract store, hopefully it wouldnt be too difficult to hook this up to postgres. will speed up development to have a common store interface

packages/indexer/src/webhooks/store.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  async *keys(): AsyncIterableIterator<string> {
+    for (const key of this.map.keys()) {
+      yield key;
+    }
+  }
+}
+
+export class RedisStore<V> implements AsyncStore<V> {
```
daywiss, line 55: redis store if we want to use it

packages/indexer/src/webhooks/utils.ts
```diff
@@ -0,0 +1,17 @@
+export async function post(url: string, data: any): Promise<void> {
```
daywiss, line 1: what we use to post to webhook

packages/indexer/src/webhooks/webhooks.ts
```diff
@@ -0,0 +1,93 @@
...
+        urlDomain === "localhost" || urlDomain.startsWith("127.");
+      if (!isDevDomain) {
+        const isDomainValid = clients.some((client) =>
+          client.domains.includes(urlDomain),
+        );
+        assert(
+          isDomainValid,
+          "The base URL of the provided webhook does not match any of the client domains",
+        );
+      }
```
daywiss, line 74: this was my attempt at gaurding a webhook for a domain, but this may need to be tested/rethought

packages/indexer/src/webhooks/express.ts
```diff
@@ -0,0 +1,83 @@
...
+  deps: Dependencies,
+): express.Application {
+  const app = express();
+  const port = config.port ?? 3000;
+
+  app.use(express.json());
+  app.use(bearerToken());
+
+  app.post(
+    "/webhook",
```
daywiss, line 36: integrator register a new webhook

amateima (changes requested):
Looks good. Left a few comments.

packages/webhooks/src/router.ts
```diff
@@ -0,0 +1,67 @@
...
+  type: ss.string(),
+  url: ss.string(),
+  filter: ss.record(ss.string(), ss.any()),
+});
+const UnregisterParams = ss.object({
+  type: ss.string(),
+  id: ss.string(),
+});
+
+export function WebhookRouter(deps: Dependencies): express.Router {
```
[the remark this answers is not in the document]
daywiss: >following our conversation on Friday, the best for now would be to remove the express logic from the webhooks package and declare a /webhooks/register endpoint in the indexer API that (1) saves the webhook request in the DB

This logic has been updated to remove references to express app, and only return a router, which is very portable.  The reason i think we should at least return a router, is because there needs to be type validation on the routes, and you dont want the consumer of the webhook library to need to know the types the library expects to correctly validate.

> P.S I asked Matt if the initial status should be sent to the webhook request URL or if we can attach it to the response body of the /register endpoint. But must likely it will be option (1)

would highly prefer all data updates go through webhook calls. it simplifies both our logic and consumer logic.

packages/webhooks/src/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  | string
+  | number
+  | boolean
+  | { [x: string]: JSONValue }
+  | Array<JSONValue>
+  | null;
+
+export type NotificationPayload = {
+  url: string;
+  event: JSONValue;
```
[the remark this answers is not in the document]
daywiss: ive renamed this to "data"

packages/webhooks/src/notifier.ts
```diff
@@ -0,0 +1,102 @@
...
+};
+
+export class BaseNotifier {
+  private pending: AsyncStore<Notification>;
+  private completed: AsyncStore<Notification>;
+  private logger: Logger;
+
+  constructor(private deps: Dependencies) {
+    this.pending = deps.pending;
+    this.completed = deps.completed;
```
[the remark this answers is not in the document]
daywiss: removed any stateful notifications

packages/webhooks/src/types.ts
```diff
@@ -0,0 +1,39 @@
+import * as ss from "superstruct";
+
+export interface Webhook {
+  id: string;
+  url: string;
+  filter: string;
+}
```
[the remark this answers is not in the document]
daywiss: renamed this

packages/webhooks/src/database/webhookClientRepository.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+    await this.store.set(client.id, client);
+  }
+
+  public async unregisterClient(clientId: string): Promise<void> {
+    if (!(await this.store.has(clientId))) {
+      throw new Error(`Client with id ${clientId} does not exist.`);
+    }
+    await this.store.delete(clientId);
+  }
```
james-a-morris, line 26:
```suggestion
  public unregisterClient(clientId: string): Promise<void> {
    if (!(await this.store.has(clientId))) {
      throw new Error(`Client with id ${clientId} does not exist.`);
    }
    return this.store.delete(clientId);
  }
```

packages/webhooks/src/database/webhookClientRepository.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  public async getClient(clientId: string): Promise<WebhookClient | undefined> {
```
james-a-morris, line 28:
```suggestion
  public getClient(clientId: string): WebhookClient | undefined {
```

packages/webhooks/src/database/webhookClientRepository.ts
james-a-morris, line 1: a lot of these functions may not need to be async

packages/webhooks/src/store.ts
```diff
@@ -0,0 +1,107 @@
...
+    const keys = await this.client.keys("*");
+    for (const key of keys) {
+      const value = await this.get(key);
+      if (value !== undefined) {
+        yield value;
+      }
+    }
+  }
+
+  async *entries(): AsyncIterableIterator<[string, V]> {
```
james-a-morris, line 91: OOC what is this *entries invocation?

packages/webhooks/src/database/webhookClientRepository.ts
daywiss: these need async, the stores are async cuz they will be calling a database

packages/webhooks/src/database/webhookClientRepository.ts
daywiss: the store functions are all async, even though right now they are using a memory map, eventually will plugin a real database

amateima (approved):

PR #103, opened 2024-11-15
fix: replace bulk upserts with inserts/updates if needed

amateima:
- new [BlockchainEventRepository](https://github.com/across-protocol/indexer/pull/103/files#diff-c7565999fde564c1bde14abd561f7bd626a30813513746d1d9bd03163d7366a0) which is intended to be extended by any repository class dealing with events entities.
- [BlockchainEventRepository.saveAndHandleFinalisation()]() is a function which takes event data and performs a save operation depending on the passed `uniqueKeys` and `comparisonKeys`. The result of this save operation can be  (1) **Nothing** if the data is identical to what's stored in the database , (2) **Inserted**, if the unique keys are not present in the DB, (3) **Finalised**, if the finalised field was the only one that changed, (4) **Updated**, if any of the data fields were changed, (5) **UpdatedAndFinalised**, if both the finalised field and other fields were changed
- [filterSaveQueryResults](https://github.com/across-protocol/indexer/pull/103/files#diff-c7565999fde564c1bde14abd561f7bd626a30813513746d1d9bd03163d7366a0R6-R14) takes a list of `SaveQueryResult[]` and filters the data based on a `SaveQueryResultType`, so for example, we can know which events were inserted in the DB for the first time
- `SpokePoolProcessor` functions were updated to perform operations only when necessary (mostly on `Inserted` and `Updated` entities) 
- [TODOs marks](https://github.com/across-protocol/indexer/pull/103/files#diff-b5925a4c81447e8689bf4dbb44dd7419585f80a247caf816407d515b21d26befR40-R72) where webhooks should be triggered

Across Indexer (15)

### DIFF
--- a/packages/indexer-database/src/utils/BlockchainEventRepository.ts
+++ b/packages/indexer-database/src/utils/BlockchainEventRepository.ts
@@ -1,0 +1,121 @@
+import { DataSource, EntityTarget, ObjectLiteral } from "typeorm";
+import winston from "winston";
+
+import { SaveQueryResultType, SaveQueryResult } from "../model";
+
+export function filterSaveQueryResults<Entity extends ObjectLiteral>(
+  results: SaveQueryResult<Entity>[],
+  type: SaveQueryResultType,
+) {
+  return results
+    .filter((result) => result.result === type)
+    .map((result) => result.data)
+    .filter((data) => data !== undefined);
+}
+
+export class BlockchainEventRepository {
+  constructor(
+    protected postgres: DataSource,
+    protected logger: winston.Logger,
+  ) {}
+
+  /**
+   * Saves the entities to the database.
+   * @param entity - The entity to save.
+   * @param data - The data to save.
+   * @param uniqueKeys
+   * The unique keys to check for. It is recommended these keys to be indexed columns, so that the query is faster.
+   * @param comparisonKeys - The keys to compare for changes.
+   */
+  protected async saveAndHandleFinalisationBatch<Entity extends ObjectLiteral>(
+    entity: EntityTarget<Entity>,
+    data: Partial<Entity>[],
+    uniqueKeys: (keyof Entity)[],
+    comparisonKeys: (keyof Entity)[],
+  ): Promise<SaveQueryResult<Entity>[]> {
+    return Promise.all(
+      data.map((dataItem) =>
+        this.saveAndHandleFinalisation(
+          entity,
+          dataItem,
+          uniqueKeys,
+          comparisonKeys,
+        ),
+      ),
+    );
+  }
+
+  /**
+   * Saves the entity to the database.
+   * @param entity - The entity to save.
+   * @param data - The data to save.
+   * @param uniqueKeys
+   * The unique keys to check for. It is recommended these keys to be indexed columns, so that the query is faster.
+   * @param comparisonKeys - The keys to compare for changes.
+   */
+  protected async saveAndHandleFinalisation<Entity extends ObjectLiteral>(
+    entity: EntityTarget<Entity>,
+    data: Partial<Entity>,
+    uniqueKeys: (keyof Entity)[],
+    comparisonKeys: (keyof Entity)[],
+  ): Promise<SaveQueryResult<Entity>> {
+    const where = uniqueKeys.reduce(
+      (acc, key) => {
+        acc[key] = data[key];
+        return acc;
+      },
+      {} as Record<keyof Entity, any>,
+    );
+    const dbEntity = await this.postgres
+      .getRepository(entity)
+      .findOne({ where });
+
+    if (!dbEntity) {
+      await this.postgres.getRepository(entity).insert(data);
+      return {
+        data: (await this.postgres
+          .getRepository(entity)
+          .findOne({ where })) as Entity,
+        result: SaveQueryResultType.Inserted,
+      };
+    }
+
+    // Check if the any of values of the comparison keys have changed
+    const isChanged = comparisonKeys.some((key) => data[key] !== dbEntity[key]);
+    // Check if the data moved in finalised state
+    const isFinalisedChanged = data.finalised && !dbEntity.finalised;
+
+    if (isChanged) {
+      await this.postgres.getRepository(entity).update(where, data);
+      if (isFinalisedChanged) {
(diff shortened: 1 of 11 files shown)


packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -42,14 +48,14 @@ export class SpokePoolRepository extends dbUtils.BaseRepository {
         finalised: event.blockNumber <= lastFinalisedBlock,
       };
     });
-    const chunkedEvents = across.utils.chunk(formattedEvents, this.chunkSize);
+    const chunkedEvents = across.utils.chunk(formattedEvents, 100);
```
melisaguevara, line 51: should we keep `this.chunkSize` here?

packages/indexer-database/src/utils/BlockchainEventRepository.ts
```diff
@@ -0,0 +1,121 @@
...
+      return {
+        data: (await this.postgres
+          .getRepository(entity)
+          .findOne({ where })) as Entity,
+        result: SaveQueryResultType.Inserted,
+      };
+    }
+
+    // Check if the any of values of the comparison keys have changed
+    const isChanged = comparisonKeys.some((key) => data[key] !== dbEntity[key]);
```
daywiss, line 84: how do you decide what props to compare?

packages/indexer-database/src/utils/BlockchainEventRepository.ts
```diff
@@ -0,0 +1,121 @@
...
+    // Check if the data moved in finalised state
+    const isFinalisedChanged = data.finalised && !dbEntity.finalised;
+
+    if (isChanged) {
+      await this.postgres.getRepository(entity).update(where, data);
+      if (isFinalisedChanged) {
+        return {
+          data: (await this.postgres
+            .getRepository(entity)
+            .findOne({ where })) as Entity,
```
daywiss, line 94: why requery this? shouldnt it be the same as "data"

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -144,11 +150,11 @@ export class SpokePoolRepository extends dbUtils.BaseRepository {
     const chunkedEvents = across.utils.chunk(formattedEvents, this.chunkSize);
     const savedEvents = await Promise.all(
       chunkedEvents.map((eventsChunk) =>
-        this.insertWithFinalisationCheck(
+        this.saveAndHandleFinalisationBatch<entities.RequestedSpeedUpV3Deposit>(
           entities.RequestedSpeedUpV3Deposit,
           eventsChunk,
           ["depositId", "originChainId", "transactionHash", "logIndex"],
-          lastFinalisedBlock,
+          ["transactionHash"],
```
melisaguevara, line 157: Should we use a different set for uniqueKeys here to avoid having `transactionHash` both in `uniqueKeys` and `comparisonKeys`?

daywiss (approved):
looks good, mainly just questions!

melisaguevara (commented):
looks good! I left some questions

packages/indexer-database/src/utils/BlockchainEventRepository.ts
```diff
@@ -0,0 +1,121 @@
...
+    const where = uniqueKeys.reduce(
+      (acc, key) => {
+        acc[key] = data[key];
+        return acc;
+      },
+      {} as Record<keyof Entity, any>,
+    );
+    const dbEntity = await this.postgres
+      .getRepository(entity)
+      .findOne({ where });
```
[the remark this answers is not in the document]
amateima: most of the events are identifiable by a pair of fields. For example, in case of a re-org, the only constant in a deposit event are `depositId` and `originChainId` fields, so you need query by these 2 fields

packages/indexer-database/src/utils/BlockchainEventRepository.ts
amateima: This logic will be specific for each event individually. Basically for each event you need to identify the fields that are unique in case of re-orgs and the fields that will be changed and needs to be compared

packages/indexer-database/src/utils/BlockchainEventRepository.ts
amateima: I was thinking for this function to return the db entity because it contains the generated fields too, such as the incremental id

packages/indexer-database/src/utils/BlockchainEventRepository.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    if (!dbEntity) {
+      await this.postgres.getRepository(entity).insert(data);
```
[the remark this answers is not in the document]
amateima: Done https://github.com/across-protocol/indexer/pull/103/commits/8a16b88e4f35e711d5354484e4e2f8ee41b9ba67

packages/indexer/src/database/SpokePoolRepository.ts
amateima: Done, but I lowered the default value of the `chunkSize`

packages/indexer/src/database/SpokePoolRepository.ts
amateima: This is intended here because unfortunately there's no way to identify uniquely a speedup event in case of re-orgs :(

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -95,20 +108,12 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
...
-
-    if (!fetchIntegratorIdSync) {
-      await this.publishIntegratorIdMessages(storedEvents.deposits);
-    }
-
+    const newInsertedDeposits = indexerDatabaseUtils.filterSaveQueryResults(
+      storedEvents.deposits,
+      SaveQueryResultType.Inserted,
+    );
+    await this.updateNewDepositsWithIntegratorId(newInsertedDeposits);
```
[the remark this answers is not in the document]
amateima: I wanted to give it a try again and see how much time it adds in production if we do this sync

PR #109, opened 2024-11-20
improve(spokePoolProcessor): improved logic for setting refunded status

melisaguevara:
Previously, we were querying the expired deposits and then querying its possible refunds. Now, we are getting from the database only the deposits that might be refunded based on the bundle events that are stored in the db.

This PR also fixes the bug we were having before (Cannot read properties of undefined, reading `relayerRefundRoot`). The issue was fixed by including the inner join with the bundle table when querying bundle events.

### DIFF
--- a/packages/indexer/src/services/spokePoolProcessor.ts
+++ b/packages/indexer/src/services/spokePoolProcessor.ts
@@ -173,38 +173,28 @@ export class SpokePoolProcessor {
       at: "SpokePoolProcessor#updateRefundedDepositsStatus",
       message: `Updating status for refunded deposits`,
     });
-    const relayHashInfoRepository = this.postgres.getRepository(
-      entities.RelayHashInfo,
-    );
     const bundleEventsRepository = this.postgres.getRepository(
       entities.BundleEvent,
     );
-
-    const expiredDeposits = await relayHashInfoRepository
-      .createQueryBuilder("rhi")
-      .select("rhi.relayHash")
-      .where("rhi.status = :expired", { expired: entities.RelayStatus.Expired })
+    const refundEvents = await bundleEventsRepository
+      .createQueryBuilder("be")
+      .innerJoinAndSelect("be.bundle", "bundle")
+      .innerJoin(entities.RelayHashInfo, "rhi", "be.relayHash = rhi.relayHash")
+      .where("be.type = :expiredDeposit", {
+        expiredDeposit: entities.BundleEventType.ExpiredDeposit,
+      })
+      .andWhere("rhi.status = :expired", {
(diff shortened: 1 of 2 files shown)


packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -173,38 +173,28 @@ export class SpokePoolProcessor {
...
+    const refundEvents = await bundleEventsRepository
+      .createQueryBuilder("be")
+      .innerJoinAndSelect("be.bundle", "bundle")
+      .innerJoin(entities.RelayHashInfo, "rhi", "be.relayHash = rhi.relayHash")
+      .where("be.type = :expiredDeposit", {
+        expiredDeposit: entities.BundleEventType.ExpiredDeposit,
+      })
+      .andWhere("rhi.status = :expired", {
+        expired: entities.RelayStatus.Expired,
+      })
```
james-a-morris, line 188: Does this generate a list of refunds without needing to do it iteratively in the loop below?

melisaguevara: Yes exactly

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -235,15 +225,16 @@ export class SpokePoolProcessor {
...
+        .getRepository(entities.RelayHashInfo)
         .createQueryBuilder()
         .update()
         .set({
           status: entities.RelayStatus.Refunded,
           depositRefundTxHash: executedRelayerRefundRootEvent.transactionHash,
         })
         .where("relayHash = :relayHash", {
-          relayHash: expiredDeposit.relayHash,
+          relayHash: refundEvent.relayHash,
```
amateima, line 237: was the previous used value incorrect?

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -235,15 +225,16 @@ export class SpokePoolProcessor {
...
         })
         .execute();
```
amateima, line 239: Looks good! Only one improvement suggestion. We need to emit webhook notifications from here with a payload like 
```
{ depositId, originChainId, status, depositTxHash }
```

Any thoughts what's the best way to do it?

melisaguevara: I think this should help to integrate the webhook function next [48f3d16](https://github.com/across-protocol/indexer/pull/109/commits/48f3d1618f6233bb3bcd3b6e14054c3a73963a42)

amateima (approved):

PR #113, opened 2024-11-22
feat(errors): add error class and functions

james-a-morris:
Initial PR to integrate a uniform data payload in the API

### DIFF
--- a/packages/error-handling/src/errors/AssertError.ts
+++ b/packages/error-handling/src/errors/AssertError.ts
@@ -1,0 +1,7 @@
+import { IndexerError } from "./IndexerError";
+
+export class AssertError extends IndexerError {
+  constructor(message: string) {
+    super(AssertError.name, message);
+  }
+}
--- a/packages/error-handling/src/errors/IndexerError.ts
+++ b/packages/error-handling/src/errors/IndexerError.ts
@@ -1,0 +1,25 @@
+/**
+ * Generic Error class that should be used in the Indexer to
+ * provide common error patterns to log
+ */
+export abstract class IndexerError extends Error {
+  constructor(
+    private readonly errorName: string,
+    private readonly errorMessage: string,
+    private readonly errorData?: Record<string, string>,
+  ) {
+    super(errorMessage);
+  }
+
+  /**
+   * A function used by `JSON.stringify` to specify which data will be serialized
+   * @returns A formatted JSON
+   */
+  public toJSON(): Record<string, unknown> {
+    return {
+      error: this.errorName,
+      message: this.errorMessage,
+      data: this.errorData,
+    };
+  }
+}
--- a/packages/error-handling/src/errors/IndexerHTTPError.ts
+++ b/packages/error-handling/src/errors/IndexerHTTPError.ts
@@ -1,0 +1,29 @@
+import { StatusCodes } from "http-status-codes";
+import { IndexerError } from "./IndexerError";
+
+/**
+ * Used to distinguish a similar design pattern as {@link IndexerError} but with
+ * additional HTTP context
+ * @see {@link IndexerError}
+ */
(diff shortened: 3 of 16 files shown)


packages/error-handling/src/errors/AssertError.ts
```diff
@@ -0,0 +1,7 @@
+import { IndexerError } from "./IndexerError";
+
+export class AssertError extends IndexerError {
+  constructor(message: string) {
+    super(AssertError.name, message);
```
amateima, line 5: does this mean that the HTTP response body will be 
```
{
  error: "AssertError",
  message: message
}
```
?

packages/error-handling/src/errors/IndexerHTTPError.ts
```diff
@@ -0,0 +1,29 @@
...
+import { IndexerError } from "./IndexerError";
+
+/**
+ * Used to distinguish a similar design pattern as {@link IndexerError} but with
+ * additional HTTP context
+ * @see {@link IndexerError}
+ */
+export abstract class IndexerHTTPError extends IndexerError {
+  constructor(
+    private readonly httpStatusCode: StatusCodes,
```
amateima, line 11: should we make this optional and set a default code? 400 or 500

packages/error-handling/src/errors/IndexerHTTPError.ts
```diff
@@ -0,0 +1,29 @@
+import { StatusCodes } from "http-status-codes";
```
amateima, line 1: maybe we should export these too so that the indexer-api would not have to install this package

packages/error-handling/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./errors";
+export * as utils from "./utils";
```
amateima, line 2: let's put all the exports in a single place
```suggestion
export * from "./utils";
```

amateima (changes requested):
Don't forget to add the package to the Dockerfile

daywiss (approved):

packages/error-handling/src/errors/IndexerHTTPError.ts
james-a-morris: https://github.com/across-protocol/indexer/pull/113/commits/a8d11c18598a6a8ab2fdba866f3f30f205ba9e3a

packages/error-handling/src/errors/IndexerHTTPError.ts
james-a-morris: I think it should be intentional otherwise we'd want to use `IndexerError`

packages/error-handling/src/errors/IndexerError.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Generic Error class that should be used in the Indexer to
+ * provide common error patterns to log
+ */
+export abstract class IndexerError extends Error {
```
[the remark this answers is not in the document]
james-a-morris: Quite possibly - I will leave it to @alexandrumatei36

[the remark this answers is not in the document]
amateima: @james-a-morris we can do a quick renaming. I agree with David that we should design these packages as they would be used in the future in other repos/apps too

amateima (approved):
This is great work! It can be merged after renaming the error class

packages/error-handling/src/errors/IndexerError.ts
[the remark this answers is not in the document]
james-a-morris: I'd prefer to avoid naming it just `Error` when dropping the `Indexer`. How about `FormattedError`?

james-a-morris:
> This is great work! It can be merged after renaming the error class

resolved offline - we'll follow in post

PR #114, opened 2024-11-22
feat: integrate webhooks in the indexer

amateima:
- integrate webhooks package in the indexer
- trigger initial notification after webhook request registration using `WebhookRequest` queue
- make apiKey required


packages/webhooks/src/utils.ts
```diff
@@ -1,10 +1,11 @@
 import { NotificationPayload } from "./types";
 export async function post(params: NotificationPayload): Promise<void> {
-  const { url, data } = params;
+  const { url, data, apiKey } = params;
   const response = await fetch(url, {
     method: "POST",
     headers: {
       "Content-Type": "application/json",
+      Authorization: `Bearer ${apiKey}`,
```
james-a-morris, line 8: OOC - why did we go with fetch vs `axios`?

packages/webhooks/src/types.ts
```diff
@@ -6,7 +6,7 @@ export interface IEventProcessor {
     id: string,
     url: string,
     params: JSONValue,
-    clientId?: string,
+    clientId: number,
```
james-a-morris, line 9: Could we not represent a number as a numeric string?

packages/webhooks/src/router.ts
```diff
@@ -27,17 +27,20 @@ export function WebhookRouter(deps: Dependencies): express.Router {
...
     "/webhook",
     async (
-      req: express.Request & { token?: string },
+      req: express.Request,
       res: express.Response,
       next: express.NextFunction,
     ) => {
       try {
         const parsedBody = RegistrationParams.create(req.body);
+        if (!req.token) {
```
james-a-morris, line 36: If we created a local token variable `const token = req.token` and checked it directly then we wouldn't need to typecast below.

packages/webhooks/src/database/webhookClientRepository.ts
james-a-morris, line 1: OOC what if two or more separate clients ask to send data to the same callback url?

packages/webhooks/src/adapter/messaging/WebhooksQueuesService.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  public async publishMessage<T>(
+    queue: WebhooksQueues,
+    message: T,
+    options: JobsOptions = {},
+  ) {
+    const q = this.queues[queue];
+    if (q) {
+      await q.add(queue, message, options);
+    }
```
james-a-morris, line 37:
```suggestion
      await this.queues[queue]?.add(queue, message, options);
```

james-a-morris: We may want to use this in multiple spots

james-a-morris (commented):
few nits + few questions. Looks great

packages/webhooks/src/utils.ts
amateima: 🤷‍♂️ this package was started with `fetch` so I went with it too. Also, I am a fan of using as few external packages as possible

amateima: `axios` has a nicer syntax tho...

packages/webhooks/src/types.ts
amateima: not sure I understand the question. This change happened because I updated the `WebhookClient` primary key from a string id computed by us to an incremental number managed by postgres

packages/webhooks/src/database/webhookClientRepository.ts
amateima: I don't think that's possible because we check the callback url agains the domains list from the `WebhookClient` and that table is managed manually by us for now

packages/webhooks/src/adapter/messaging/WebhooksQueuesService.ts
amateima: @james-a-morris I plan to add a `warn | error` log here on else, that's why I used if

packages/webhooks/src/router.ts
amateima: fixed

packages/webhooks/src/adapter/messaging/WebhookRequestWorker.ts
```diff
@@ -0,0 +1,84 @@
...
+            error,
+          });
+          throw error;
+        }
+      },
+      { connection: this.redis, concurrency: 10 },
+    );
+  }
+
+  private async run(webhookRequestJob: WebhookRequestQueueJob) {
```
daywiss, line 52: really against putting this business logic here, im happy to take a crack at refactoring this in a future pr though.
i think what id do is basically pass in the eventprocessor manager in the webhookrequestworker constructor, then assume whenever "run" is called, its a webhook registration, and call "register" on the eventprocessormanager, which will call the correct event processor and put the logic for pulling in initial state and calling webhook in there. 

its a subtle change, but means that we could remove this worker if everythign resides on the same process

packages/webhooks/src/eventProcessorManager.ts
```diff
@@ -101,7 +113,7 @@ export class EventProcessorManager {
     this.logger.debug(
       `Attempting to unregister webhook of type: ${params.type} with ID: ${params.id}`,
     );
-    const webhook = this.getEventProcessor(params.type);
+    const webhook = this.getEventProcessor(params.type as WebhookTypes);
```
daywiss, line 116: could make a function to assert this

packages/webhooks/src/eventProcessors/depositStatus.ts
```diff
@@ -24,56 +26,90 @@ export type DepositStatusFilter = ss.Infer<typeof DepositStatusFilter>;
...
+
     //TODO: unregister any hooks where event has reached terminal state
     await Promise.all(
-      hooks.map((hook) => {
-        this.notify({
-          url: hook.url,
-          data: event,
-        });
+      webhookRequests.map((hook) => {
+        const client = clientsMap[hook.clientId];
```
daywiss, line 73: theres a lot of work above just to get this cache, not sure why u dont just make this function async(you are already using promise.all even though function is currently sync), and query for the client directly here, and delete all the code from 54-69

packages/webhooks/src/types.ts
daywiss: i tihnk he means its more flexible to keep this as a string and use number.toString or parseInt to convert from and to postgres

packages/webhooks/src/eventProcessors/depositStatus.ts
amateima: I removed Promise.all(). Basically I don't want to fetch the client for each request because this will trigger many duplicated queries, so I fetch only the unique clients

packages/webhooks/src/adapter/messaging/WebhookRequestWorker.ts
amateima: I agree that this is the best approach, but what if I place this worker in the indexer package? 

> put the logic for pulling in initial state and calling webhook in there.

I do feel that the `write()` function should be called only externally and not from the webhooks package. The package should only deal with forwarding notification payloads from our systems to clients and not dealing at all from where `write()` is being called. It can be procedural code, cron jobs, Bull workers, etc. wdyt?

daywiss: we should discuss further offline, but as of now i think this is ok, we can figure out how to change later

daywiss (approved):

PR #115, opened 2024-11-25
feat: create uniform error structure

james-a-morris:


packages/error-handling/src/errors/IndexerHTTPError.ts
```diff
@@ -10,22 +10,11 @@ export { StatusCodes };
...
-  /**
-   * A function used by `JSON.stringify` to specify which data will be serialized
-   * @returns A formatted JSON
-   */
-  public toJSON(): Record<string, unknown> {
-    return {
-      statusCode: this.httpStatusCode,
-      ...super.toJSON(),
-    };
-  }
```
amateima, line 30: OOC why was this removed?

packages/indexer-api/src/error-handler.ts
```diff
@@ -0,0 +1,45 @@
...
+      : DEFAULT_STATUS;
+    res.status(httpStatus).json(err.toJSON());
+  } else if (isHttpError(err)) {
+    res.status(err.status ?? DEFAULT_STATUS).json({
+      message: err.message,
+      error: "NavigationError",
+    });
+  } else if (err instanceof StructError) {
+    res.status(StatusCodes.BAD_REQUEST).json({
+      error: "InputValidationError",
```
amateima, line 34: We can simplify this to `ValidationError`

packages/error-handling/src/errors/IndexerHTTPError.ts
james-a-morris: It would add the status code to the response code but this info is already exposed via the http status code that will get thrown.

packages/indexer-api/src/error-handler.ts
james-a-morris: https://github.com/across-protocol/indexer/pull/115/commits/d5d6c4da68894c50f5abfc787693f378176fc85d

amateima (approved):

PR #116, opened 2024-11-25
chore: replace info-level logs with debug

melisaguevara:
In order to use UMA's logger to send logs to slack, we should reduce the number of info logs in the indexer.

This PR:
- Replaces almost all info logs with debug logs
- For error-level logs, it sets the notificationPath key to "across-indexer-error" so that those are sent to a error specific channel.
- Adds the "Indexer" or "IndexerAPI" prefix to the "at" value of all logs so that we can easily filter in GCP and exclude SDK logs.

Zion PR to config slack channels: https://github.com/UMAprotocol/zion/pull/429

### DIFF
--- a/packages/indexer-api/src/main.ts
+++ b/packages/indexer-api/src/main.ts
@@ -18,8 +18,8 @@ async function initializeRedis(
 
   return new Promise<Redis>((resolve, reject) => {
     redis.on("ready", () => {
-      logger.info({
-        at: "Indexer-API",
+      logger.debug({
+        at: "IndexerAPI",
         message: "Redis connection established",
         config,
       });
@@ -28,8 +28,9 @@ async function initializeRedis(
 
     redis.on("error", (err) => {
       logger.error({
-        at: "Indexer-API",
+        at: "IndexerAPI",
         message: "Redis connection failed",
+        notificationPath: "across-indexer-error",
         error: err,
       });
       reject(err);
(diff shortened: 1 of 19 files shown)


packages/indexer-api/src/main.ts
```diff
@@ -28,8 +28,9 @@ async function initializeRedis(
 
     redis.on("error", (err) => {
       logger.error({
-        at: "Indexer-API",
+        at: "IndexerAPI",
```
james-a-morris, line 31: We may want to specify what the function is. I.e. `IndexerAPI#main`

packages/indexer/src/utils/bundleBuilderUtils.ts
```diff
@@ -34,8 +34,9 @@ export async function resolveMostRecentProposedAndExecutedBundles(
   // If no executed bundle is found, the system is in an inconsistent state
   if (!utils.isDefined(lastExecutedBundle)) {
     logger.error({
+      at: "Indexer#resolveMostRecentProposedAndExecutedBundles",
       message: "No executed bundles found",
-      at: "resolveMostRecentProposedAndExecutedBundles",
+      notificationPath: "across-indexer-error",
```
james-a-morris, line 39: ooc what is a `notificationPath` in this case?

melisaguevara: This is the key we use to send Slack messages to a specific channel.
It has to be part of the `escalationPathWebhookUrls` object in the SLACK_CONFIG env var.
Here's our config: https://github.com/UMAprotocol/zion/pull/429/files
We'll have a default channel for info and warns and other channel just for errors

packages/indexer-api/src/main.ts
melisaguevara: Done here [b59aa5c](https://github.com/across-protocol/indexer/pull/116/commits/b59aa5c0331c217349550331d1682efdc0f23077)

PR #117, opened 2024-11-27
fix(webhooks): make route handling sync, add get route

daywiss:
# motivation
Seems as if webhook route in production is not working

# changes
Unable to reproduce this locally, but perhaps the issue is with the async routes, so made this sync as well as put a "health" check get endpoint for a sanity check


daywiss:
turns out this is workign in production, just had the url slightly wrong when testing

PR #118, opened 2024-11-27
feat(webhooks): allow client registration through env

daywiss:
# motivation
we need an easy way to add and update arbitrary webhook clients

# changes
this adds env config var to regsiter new clients as well as removes domain checks to make it easier for testing

### DIFF
--- a/packages/indexer-api/src/main.ts
+++ b/packages/indexer-api/src/main.ts
@@ -87,10 +87,12 @@ export async function Main(
   const postgres = await connectToDatabase(postgresConfig, logger);
   const redisConfig = Indexer.parseRedisConfig(env);
   const redis = await initializeRedis(redisConfig, logger);
-  const webhooks = Webhooks.WebhookFactory(
+  const webhooks = await Webhooks.WebhookFactory(
     {
       enabledWebhooks: [Webhooks.WebhookTypes.DepositStatus],
       enabledWebhookRequestWorkers: false,
+      // indexer will register clients
+      clients: [],
     },
     { postgres, logger, redis },
   );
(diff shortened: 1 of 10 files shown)


packages/indexer-api/src/main.ts
```diff
@@ -87,10 +87,12 @@ export async function Main(
   const postgres = await connectToDatabase(postgresConfig, logger);
   const redisConfig = Indexer.parseRedisConfig(env);
   const redis = await initializeRedis(redisConfig, logger);
-  const webhooks = Webhooks.WebhookFactory(
+  const webhooks = await Webhooks.WebhookFactory(
```
james-a-morris, line 90: Usually a factory is a class that has a builder function on it. Is this Factory a function?

packages/indexer-database/src/entities/WebhookClient.ts
```diff
@@ -2,6 +2,7 @@ import { Entity, Column, PrimaryGeneratedColumn, Unique } from "typeorm";
 
 @Entity()
 @Unique("UK_webhook_client_api_key", ["apiKey"])
+@Unique("UK_webhook_client_name", ["name"])
```
daywiss, line 5: i wnated to make a unique constraint for name, i dont know if i did this right because i have no idea what UK_webhook means

packages/indexer-database/src/migrations/1732730161339-Webhook.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Webhook1732730161339 implements MigrationInterface {
```
daywiss, line 3: did the migration after adding uniquye name constraint, again dont know if this is correct

james-a-morris (commented):
maybe I'm a bit confused on this but shouldn't CB be able to register their own client hooks?

james-a-morris (approved):
as long as this runs the code looks fine outside of stylizations

PR #121, opened 2024-11-27
fix: check for overflow in configs

james-a-morris:
There's a possibility that if the `lastBlockFinalizedStored` is greater than the `lastFinalizedBlockOnChain` then we will have the from block be larger than the to block


james-a-morris:
Will close this out as the issue was of another origin

PR #124, opened 2024-12-02
fix: aleph zero block time

melisaguevara:
We were making requests for blocks that were not minted yet because Aleph Zero's block time is higher that the one we had set.

+2 minor improvements.

### DIFF
--- a/packages/indexer/src/services/bundles.ts
+++ b/packages/indexer/src/services/bundles.ts
@@ -82,7 +82,7 @@ function logResultOfAssignment(
   unassociatedRecordsCount: number,
   persistedRecordsCount: number,
 ): void {
-  if (unassociatedRecordsCount > 0) {
+  if (persistedRecordsCount > 0) {
     logger.debug({
       at: `Indexer#BundleEventsProcessor#assignToBundle`,
       message: "Found and associated events with bundles",
@@ -294,7 +294,11 @@ async function assignBundleRangesToProposal(
     }),
   );
   const insertResults = await dbRepository.associateBlockRangeWithBundle(
-    rangeSegments.filter((segment) => segment !== undefined).flat(),
+    rangeSegments
+      .filter(
+        (segment): segment is BlockRangeInsertType[] => segment !== undefined,
+      )
+      .flat(),
   );
   logResultOfAssignment(
     logger,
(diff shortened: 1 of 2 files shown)


packages/indexer/src/services/bundles.ts
```diff
@@ -294,7 +294,11 @@ async function assignBundleRangesToProposal(
     }),
   );
   const insertResults = await dbRepository.associateBlockRangeWithBundle(
-    rangeSegments.filter((segment) => segment !== undefined).flat(),
+    rangeSegments
+      .filter(
+        (segment): segment is BlockRangeInsertType[] => segment !== undefined,
+      )
+      .flat(),
```
melisaguevara, line 301: This was previously showing this warning: 
```
Argument of type '(BlockRangeInsertType | undefined)[]' is not assignable to parameter of
type 'BlockRangeInsertType[]'.
```

packages/indexer/src/services/bundles.ts
```diff
@@ -82,7 +82,7 @@ function logResultOfAssignment(
   unassociatedRecordsCount: number,
   persistedRecordsCount: number,
 ): void {
-  if (unassociatedRecordsCount > 0) {
+  if (persistedRecordsCount > 0) {
```
melisaguevara, line 85: This is just to avoid this log when there were no changes in the database:
```
2024-12-02 16:00:45 [[34mdebug[39m]: {
  "at": "Indexer#BundleEventsProcessor#assignToBundle",
  "message": "Found and associated events with bundles",
  "unassociatedRecordsCount": 1,
  "persistedRecordsCount": 0,
  "eventType": "BundleBlockRange",
  "bot-identifier": "NO_BOT_ID",
  "run-identifier": "77aba3dc-ac34-47ff-a9e7-2c215bfce5fa"
}
```

packages/indexer/src/services/bundles.ts
james-a-morris: Are we confident that filtering this out isn't removing data that should be present in our queries?

packages/indexer/src/services/bundles.ts
james-a-morris: Fair call +1

james-a-morris (approved):
left the one comment - i don't think it's blocking but should at least be addressed

packages/indexer/src/services/bundles.ts
melisaguevara: I only added the explicit type for the array the filtering is returning. I won't expect this to change the filtering logic we have been running

daywiss (approved):

PR #125, opened 2024-12-02
fix: modify deposit status response to be backwards compatible with scraper

melisaguevara:


melisaguevara:
@james-a-morris afaik we are not doing anything yet with the response in the FE, so there should be no blockers to merge this, right?

james-a-morris:
> @james-a-morris afaik we are not doing anything yet with the response in the FE, so there should be no blockers to merge this, right?

This will cause failures in the FE as it's currently deployed on production. We will need to carefully release this change.

melisaguevara:
> breaking change to the API that will affect the FE

@james-a-morris  Ok ok, good to know!
OOC, which part of the FE will break? Do we have to open a PR there too to support this?

james-a-morris (approved):

PR #129, opened 2024-12-05
improve(indexer): add block timestamp to depositevent data

daywiss:
# motivation
We want to store block timestamps with deposit data in order to calculate some metrics

# changes
This adds a column "blockTimestamp" to v3FundsDeposited, and queries per block range all the block timestamps and adds it to the data when storing in database

# profiling
We will use this new value and the existing "createdAt" value, which is our insertion time, for instance:
`const depositWriteTime = toSeconds(deposit.createdAt) - deposit.blockTimestamp`

### DIFF
--- a/packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
+++ b/packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
@@ -89,4 +89,7 @@ export class V3FundsDeposited {
 
   @CreateDateColumn()
   createdAt: Date;
+
+  @Column({ nullable: true })
+  blockTimestamp?: number;
 }
--- a/packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
+++ b/packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
@@ -123,16 +123,15 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
     return block.timestamp;
   }
 
-  private async getBlockTimesByRange(
-    blockRange: BlockRange,
+  private async getBlockTimes(
+    blockNumbers: number[],
   ): Promise<Record<number, number>> {
     const blockTimes: Record<number, number> = {};
-    for (
-      let blockNumber = blockRange.from;
-      blockNumber <= blockRange.to;
-      blockNumber++
-    ) {
-      blockTimes[blockNumber] = await this.getBlockTime(blockNumber);
+    for (const blockNumber of blockNumbers) {
+      // ensure we dont query more than we need to
+      if (blockTimes[blockNumber] === undefined) {
+        blockTimes[blockNumber] = await this.getBlockTime(blockNumber);
+      }
     }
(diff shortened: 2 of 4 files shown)


packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
```diff
@@ -89,4 +89,7 @@ export class V3FundsDeposited {
 
   @CreateDateColumn()
   createdAt: Date;
+
+  @Column({ nullable: true })
+  blockTimestamp?: number;
```
james-a-morris, line 94: Is this something we will eventually backfill? Could be a good candidate to make required with the prod backfilling

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -46,6 +47,7 @@ export class SpokePoolRepository extends dbUtils.BlockchainEventRepository {
         ...this.formatRelayData(event),
         quoteTimestamp: new Date(event.quoteTimestamp * 1000),
         finalised: event.blockNumber <= lastFinalisedBlock,
+        blockTimestamp: blockTimes[event.blockNumber],
```
melisaguevara, line 50: What do you think about considering quoteTimestamp as deposit date? I know there might be a tiny difference with the actual block number though.

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -114,6 +115,27 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
...
+    const block = await this.provider.getBlock(blockNumber);
+    if (!block) {
+      throw new Error(`Block with number ${blockNumber} not found`);
+    }
+    return block.timestamp;
+  }
+
+  private async getBlockTimesByRange(
+    blockRange: BlockRange,
+  ): Promise<Record<number, number>> {
```
amateima, line 128: I think there's room for optimisation here. Instead of passing the whole block range, we can pass here a list with only the blocks we got deposit events for. 
This will avoid fetching thousands of blocks with no events during backfilling

packages/indexer/src/database/SpokePoolRepository.ts
amateima: I think the most accurate will be to track the time when the user's tx got mined

amateima (changes requested):

packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
daywiss: im not totally sure how this works by making it required, i wanted to avoid having to rerun indexing to get this data, and allow it optionally going forward

daywiss:
@amateima I made 2 changes:
1. optimized the block timestamp query to only care about blocks where we have deposit events
2. log out the time difference directly after saving, so we can use datadog immediately

amateima (approved):

PR #130, opened 2024-12-05
fix: chain ids mistakenly included in proposal entity

melisaguevara:
For a bundle which block ranges are:

[21181520, 127967611, 64250084, 977154, 274166888, 48874544, 22372326, 12024812, 15683208, 8402604, 11362092, 11072547, 9674454, 22420080, 6099180]

We are seting chainIds to:

[1, 10, 137, 288, 42161, 324, 8453, 59144, 34443, 1135, 81457, 534352, 690, 7777777, 480, 41455]

But 41455 shouldn't be part of it yet.

This is causing we trying to update a spoke pool with this config

```
searchConfig: {
  fromBlock: "NaN"
  maxBlockLookBack: 10000
  toBlock: "NaN"
}
```

Proposal chain ids are mistakenly configured every time a new chain is added.


amateima (approved):
Good catch

PR #138, opened 2024-12-18
feat(indexer): add coingecko price collection from tokens against usd

daywiss:
# motivation
we want to start collecting price data in the indexer

# changes
This dynamically looks up prices based on fill events, passing them to a worker to then fetch and/or cache the price at the deposit time


packages/indexer/src/parseEnv.ts
```diff
@@ -193,6 +196,9 @@ export function envToConfig(env: Env): Config {
     enabledWebhookRequestWorkers: true,
     clients: parseWebhookClientsFromString(env.WEBHOOK_CLIENTS ?? "[]"),
   };
+  const coingeckoSymbols = parseArray(env.COINGECKO_SYMBOLS).map((symbol) =>
```
amateima, line 199: We should document this ENV value in the README and specify the format of it (list, object, etc)

amateima (changes requested):

packages/indexer-database/src/entities/HistoricPrice.ts
```diff
@@ -0,0 +1,35 @@
...
+  // bear in mind we are using coingecko symbols directly here, for all intents and purposes this is coingecko historic market price
+  @Column()
+  baseCurrency: string;
+
+  @Column({ default: "usd" })
+  quoteCurrency: string;
+
+  // yyyy-LL-dd
+  @Column()
+  date: string;
```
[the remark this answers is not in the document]
daywiss: i think the complexity is that we only care about 1 price per day, which means deterministically calculating the "day" based on a js Date or timestamp becomes more involved. Ill add the logic and u can see if thats what you want

packages/indexer/src/services/priceProcessor.ts
```diff
@@ -0,0 +1,82 @@
...
+        // wasnt able to get a price
+        if (price === undefined) {
+          this.deps.logger.error(
+            `Unable to find ${quoteCurrency} for ${symbol}`,
+          );
+          return;
+        }
+        await historicPriceRepository.insert({
+          date: dbFormattedDate,
+          baseCurrency: symbol,
```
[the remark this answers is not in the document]
daywiss: this is currently an enum, so if anything is wrong we can fix it in the enum

packages/indexer-database/packages/indexer-database/src/migrations/1733941169940-HistoricPrice.ts
```diff
@@ -0,0 +1,14 @@
...
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`CREATE TABLE "historic_market_price" ("id" SERIAL NOT NULL, "baseCurrency" character varying NOT NULL, "quoteCurrency" character varying NOT NULL DEFAULT 'usd', "date" date NOT NULL, "price" numeric NOT NULL, "createdAt" TIMESTAMP NOT NULL DEFAULT now(), CONSTRAINT "UK_historic_price_baseCurrency_quoteCurrency_date" UNIQUE ("baseCurrency", "quoteCurrency", "date"), CONSTRAINT "PK_b0a22436b47e742187aa7408561" PRIMARY KEY ("id"))`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`DROP TABLE "historic_market_price"`);
+    }
+
+}
```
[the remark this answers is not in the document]
daywiss: yea this was a mistake. wil fix

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,172 @@
...
+  originChainId: number;
+};
+
+/**
+ * This worker listens to the `IntegratorId` queue and processes each job by:
+ * - Retrieving the deposit information from the database based on the provided relay hash.
+ * - Checking if the deposit record already has an integrator ID.
+ * - If the integrator ID is not set, the worker fetches it from the transaction data.
+ * - If found, the integrator ID is saved back into the deposit record in the database.
+ */
```
amateima: These comments should be updated or removed

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,172 @@
...
+export class PriceWorker {
+  public worker: Worker;
```
amateima, line 30: Is this property needed to be public?

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,172 @@
...
+    const { depositId, originChainId } = params;
+    const relayHashInfoRepository = this.postgres.getRepository(
+      entities.RelayHashInfo,
+    );
+    const depositRepository = this.postgres.getRepository(
+      entities.V3FundsDeposited,
+    );
+    const historicPriceRepository = this.postgres.getRepository(
+      entities.HistoricPrice,
+    );
```
amateima, line 70: Would make sense to make the repository variables to be class properties and set them in the constructor? This way we would reduce the scope of `run()` function by not having it to deal with instantiating dependencies

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,172 @@
...
+        ...params,
+      });
+      return;
+    }
+    const priceTime = yesterday(blockTime);
+    const quoteCurrency = "usd";
+    const baseTokenInfo = findTokenByAddress(
+      relayHashInfo.fillEvent.outputToken,
+      relayHashInfo.destinationChainId,
+    );
```
amateima, line 102: I see that `findTokenByAddress()` function works on a list of tokens that filters out the tokens for which we don't have the CoinGecko symbol defined in the indexer. I don't think this is necessary. 
The only condition we need to check is if the token is part of the supported tokens specified in the `@across-protocol/constants`. If we do this, we also don't need the CoinGecko env anymore

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,172 @@
...
+      this.logger.error({
+        at: "PriceWorker",
+        message: "Deposit block time not found for relay hash info",
+        ...params,
+      });
+      return;
+    }
+    const priceTime = yesterday(blockTime);
+    const quoteCurrency = "usd";
+    const baseTokenInfo = findTokenByAddress(
```
amateima, line 99: We need to fetch price for input token too

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,175 @@
...
+ * This worker listens to the `PriceQuery` queue and processes each job by:
+ * - Retrieving the deposit and relay hash information from the database using the deposit ID and origin chain ID.
+ * - Verifying the existence of the relay hash info and deposit records.
+ * - Determining the block time from the relay hash info and calculating the price time as the previous day's timestamp.
+ * - Identifying the base currency using the output token and destination chain ID.
+ * - Checking if a historic price for the base currency and quote currency (USD) already exists in the database.
+ * - If not, fetching the historic price from Coingecko and inserting it into the database.
+ * - Logging errors and information at various stages of the process.
+ */
+export class PriceWorker {
```
amateima, line 29: The way errors are handled in this worker should be reviewed and changed accordingly. The only situation in which the message should be discarded and execution should stop is the case when the deposit with a particular `depositId` and `originChainId` doesn't exist. Otherwise, we should log the error (as we do) **and** throw an error, so that the worker retries the message

amateima (changes requested):

packages/error-handling/src/utils/assert.ts
```diff
@@ -8,9 +8,12 @@ import { AssertError } from "../errors/AssertError";
...
  * @throws {@link AssertError} if assert's validity fails
  */
-export function assert(value: unknown, message: string): asserts value {
+export function assert(
+  value: unknown,
+  message: string,
+): asserts value is NonNullable<unknown> {
   try {
-    return assertModule.ok(value, message);
+    return assertModule.ok(value !== null && value !== undefined, message);
```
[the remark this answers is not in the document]
daywiss: i probably dont need to do this

amateima (changes requested):
I haven't looked at the code yet. I tried to test it locally, but I hit this error
Screenshot 2025-01-03 at 15 34 19

amateima (changes requested):
I am not able to run the workers successfully, see screenshot for error.

Screenshot 2025-01-06 at 13 29 49

I think the problem is at line 154 in the worker:

```
const deposit = await this.depositRepository.findOne({
      where: { depositId, originChainId },
    });
```

Is very possible for fill events to be fetched before the deposit events on the origin chain, so maybe in this case we should retry messages until the corresponding deposit event is found in the DB?

packages/indexer-database/src/entities/HistoricPrice.ts
```diff
@@ -0,0 +1,35 @@
...
+  baseCurrency: string;
+
+  @Column({ default: "usd" })
+  quoteCurrency: string;
+
+  // yyyy-LL-dd
+  @Column({ type: "date" })
+  date: Date;
+
+  @Column({ type: "decimal" })
```
james-a-morris, line 30: Is decimal needed here? Could a float be enough?

packages/indexer-database/src/entities/HistoricPrice.ts
```diff
@@ -0,0 +1,35 @@
...
+export class HistoricPrice {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  // bear in mind we are using coingecko symbols directly here, for all intents and purposes this is coingecko historic market price
+  @Column()
+  baseCurrency: string;
+
+  @Column({ default: "usd" })
+  quoteCurrency: string;
```
james-a-morris, line 24: Could be worth enumerated the expected values of these. We're unlikely to deviate from `usd` or `eth` or `azero` or `matic` as a base currency and the quote currency is unlikely to not be `usd`

packages/indexer-database/src/entities/RelayHashInfo.ts
```diff
@@ -91,6 +92,13 @@ export class RelayHashInfo {
   @CreateDateColumn()
   createdAt: Date;
 
+  @Column({ nullable: true })
+  bridgeFeeUsd: string;
+  @Column({ nullable: true })
+  inputPriceUsd: number;
+  @Column({ nullable: true })
+  outputPriceUsd: number;
```
james-a-morris, line 100: If we're willing to use `decimal` in the historic price, should we continue using it here?

packages/indexer/package.json
```diff
@@ -28,12 +28,14 @@
...
     "@types/express": "^4.17.21",
     "@types/lodash": "^4.17.7",
+    "@types/luxon": "^3.4.2",
     "bullmq": "^5.12.12",
     "ethers": "^5.7.2",
     "express": "^4.19.2",
     "express-bearer-token": "^3.0.0",
     "ioredis": "^5.4.1",
     "lodash": "^4.17.21",
+    "luxon": "^3.5.0",
```
james-a-morris, line 38: Do we need luxon for this? I think we could probably use the standard date types

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -283,6 +287,19 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
...
+  private async publishNewRelays(relays: FillWithBlock[]) {
+    const messages: PriceMessage[] = relays.map((relay) => {
+      return {
+        depositId: relay.depositId,
+        originChainId: relay.originChainId,
+      };
+    });
+    await this.indexerQueuesService.publishMessagesBulk(
+      IndexerQueues.PriceQuery,
+      IndexerQueues.PriceQuery, // Use queue name as job name
```
james-a-morris, line 299: OOC why do we repeat this parameter twice?

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -114,6 +116,8 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
     );
     await this.updateNewDepositsWithIntegratorId(newInsertedDeposits);
     await this.spokePoolProcessor.process(storedEvents);
+    // publish new relays to workers to fill in prices
+    await this.publishNewRelays(events.filledV3RelayEvents);
```
james-a-morris, line 120: Could this (and the two async calls above it) be parallelized?

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,242 @@
...
+
+    const cachedPrice = await this.historicPriceRepository.findOne({
+      where: {
+        date: priceTime,
+        baseCurrency,
+        quoteCurrency,
+      },
+    });
+    // we have this price at this time in the db
+    if (cachedPrice) return Number(cachedPrice.price);
```
james-a-morris, line 69: nit: w/out the braces it's not immediately obvious that this isn't a one-liner but rather a conditional + a return
```suggestion
    if (cachedPrice) { 
        return Number(cachedPrice.price);
    }
```

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,242 @@
...
+    // Normalize the output amount by multiplying it with its price and dividing by 10^decimals
+    // This converts the amount to a common scale based on its price
+    const normalizedOutputAmount =
+      (outputAmountBigInt * outputPriceBigInt) /
+      BigInt(Math.pow(10, outputToken.decimals));
+
+    // Calculate the bridge fee by subtracting the normalized output amount from the normalized input amount
+    // This gives the difference in value between the input and output tokens
+    return ethers.utils.formatEther(
+      normalizedInputAmount - normalizedOutputAmount,
```
james-a-morris, line 145: +1 I didn't realize that `formatEther` could process `bigint` types

packages/indexer/src/parseEnv.ts
james-a-morris, line 1: OOC why was this updated?

packages/indexer/src/utils/currencyUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+      token.address.toLowerCase() === address.toLowerCase() &&
+      token.chainId === chainId,
+  );
+  if (!result) {
+    throw new Error(
+      `Token info not found for address: ${address} on chainId: ${chainId}`,
+    );
+  }
+  return result;
+}
```
james-a-morris, line 50: We have a similar function in the SDK - maybe we should defer to this

james-a-morris (changes requested):
Could be worth reviewing: https://github.com/across-protocol/sdk/tree/master/src/coingecko

We could probably leverage this heavily to avoid code duplication. This will be helpful when we have to worry about SVM.

packages/indexer-database/src/entities/HistoricPrice.ts
daywiss: sure, im not sure what the difference is but i can change it to float

packages/error-handling/src/utils/assert.ts
```diff
@@ -8,7 +8,10 @@ import { AssertError } from "../errors/AssertError";
  * @returns An assertion of `value`.
  * @throws {@link AssertError} if assert's validity fails
  */
-export function assert(value: unknown, message: string): asserts value {
+export function assert(
+  value: unknown,
+  message: string,
+): asserts value is NonNullable<unknown> {
```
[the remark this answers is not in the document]
daywiss: after you assert something exists, typscript should be able to infer that its not null, that wasnt happening before

packages/indexer-database/src/entities/HistoricPrice.ts
```diff
@@ -0,0 +1,35 @@
+import {
+  Column,
+  CreateDateColumn,
+  Entity,
+  PrimaryGeneratedColumn,
+  Unique,
+} from "typeorm";
+
+@Entity()
```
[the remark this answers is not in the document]
daywiss: im not sure i know what this means, looking at the other entities, i dont see them being named here

packages/indexer-database/src/entities/RelayHashInfo.ts
daywiss: ill put whatever you think makes sense, im not familiar with the nuances

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
daywiss: Im just following the pattern from previous workers, but its passing the queue and then the name of the queue, is based on the enum

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
daywiss: i think so

daywiss: actually i think we want to wait for all sideeffects to complete before trying to run the job

packages/indexer-database/src/entities/HistoricPrice.ts
amateima: Not sure what float is @james-a-morris. These are the Postgres numeric types 
Screenshot 2025-01-07 at 19 41 41

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
amateima: I don't think they can be parallelized as we need to wait for deposit and fill events to be associated with relay hashes entites

packages/indexer/src/utils/currencyUtils.ts
amateima: +1 if such function exists

packages/indexer/src/messaging/priceWorker.ts
daywiss: i didnt think about that, but yea looking at the type: `type BigNumberish = BigNumber | Bytes | bigint | string | number;`

packages/indexer/src/parseEnv.ts
daywiss: originally there was some env changes made for this pr, which are no longer needed, but this was accidentally left in

daywiss: i think stylistically this is slightly better than before but i can revert if you like the other way better

packages/indexer/package.json
daywiss: basically just using it to get the time formatted to the coingecko call

packages/indexer/src/utils/currencyUtils.ts
daywiss: will need to tackle this when i get back, but was able to remove the coingecko library

packages/indexer-database/src/entities/HistoricPrice.ts
james-a-morris: Ah maybe I'm thinking of a different RDBMS

james-a-morris (commented):
Curious about https://github.com/across-protocol/indexer/pull/138#discussion_r1905680247

amateima:
@daywiss I updated the migration files as they should reflect the last version of the DB schema, not all the transformations that the DB suffered during the development

packages/indexer/src/utils/currencyUtils.ts
daywiss: i cant use the one in the sdk unfortunately, it does not return all the data i need from the token. this is what i found:
https://github.com/across-protocol/sdk/blob/master/src/utils/TokenUtils.ts#L112
which omits the coingeckoId.  will keep this in as to not block this pr for now

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,292 @@
...
+        message: errorMessage,
+        ...params,
+      });
+      throw new Error(errorMessage);
+    }
+
+    const { depositId } = relayHashInfo;
+    const deposit = await this.depositRepository.findOne({
+      where: { depositId, originChainId },
+    });
```
melisaguevara, line 192: I don't think we need this query as we are already loading the relations in line 174. We should be able to use `relayHashInfo.depositEvent`

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,292 @@
...
+      const errorMessage = "Deposit block time not found for relay hash info.";
+      this.logger.error({
+        at: "PriceWorker",
+        message: errorMessage,
+        ...params,
+      });
+      throw new Error(errorMessage);
+    }
+    const inputTokenAddress = relayHashInfo.fillEvent.inputToken;
+    const outputTokenAddress = relayHashInfo.fillEvent.outputToken;
```
melisaguevara, line 232: Are there any reasons to get this data from the fill event instead of the deposit? If not, I think we can trigger this worker for each deposit instead of for each fill.

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,292 @@
...
+    }
+    if (Number(relayHashInfo.outputPriceUsd) !== outputTokenPrice) {
+      updatedFields.outputPriceUsd = outputTokenPrice.toString();
+    }
+
+    if (Object.keys(updatedFields).length > 0) {
+      await this.relayHashInfoRepository.update(
+        { depositId, originChainId },
+        updatedFields,
+      );
```
melisaguevara, line 280: Now there could be multiple deposits matching this criteria, but I think it's still valid to use this params and update them all as all their params will match.

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,292 @@
...
+  private async run(params: PriceMessage) {
+    const { relayHash, originChainId } = params;
+
+    const relayHashInfo = await this.relayHashInfoRepository.findOne({
+      where: { relayHash, originChainId },
+      relations: {
+        depositEvent: true,
+        fillEvent: true,
+      },
+    });
```
melisaguevara, line 177: Now there could be many relayHashInfo records matching this criteria. I think if we could trigger this for each inserted deposit and send the deposit event id in the message for this worker, then we should be able to do something like:
```
const relayHashInfo = await this.relayHashInfoRepository.findOne({
      where: { depositEventId },
      relations: {
        depositEvent: true,
      },
    });
```
This would avoid for example getting one row that matches the criteria but doesn't have the deposit associated yet.

melisaguevara (commented):
Left some comments.
My main thinking is that we should trigger this worker for each inserted deposit instead of fills. Unless I'm missing something, it looks like we can get all the params we need from the deposit event.
Regarding duplicated deposits, I think we should use some unique key when sending messages to the processor. That way we can be confident we are getting the row we want to process from relayHashInfo. With the current fields (relayHash, originChainId) we can get a different row that also matches this criteria. My proposal is to use depositEventId.

packages/indexer/src/messaging/priceWorker.ts
amateima: I think it is like this in order to be able to compute the relayer gas fee in the same step

melisaguevara: mmm I see. What about deposits that don't get filled? Do we want to compute those fees?

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -332,6 +336,25 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
...
+  ) {
+    const messages: PriceMessage[] = fills
+      .map((fill) => ({
+        relayHash: fill.data?.relayHash,
+        originChainId: fill.data?.originChainId,
+      }))
+      .filter(
+        (x): x is PriceMessage =>
+          x.relayHash !== undefined && x.originChainId !== undefined,
+      );
```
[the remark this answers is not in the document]
daywiss: i just tried this, but typescript complains since it thinks data can always be undefined even after a check since `SaveQueryResult<entities.FilledV3Relay>` data can be undefined
```
type SaveQueryResult<T> = {
    data: T | undefined;
    result: SaveQueryResultType;
}
```

but im gonna change it anyway and just override typescript

packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
```diff
@@ -88,6 +88,7 @@ export class V3FundsDeposited {
   @CreateDateColumn()
   createdAt: Date;
 
+  // this has been converted from block time seconds
```
melisaguevara, line 91: What does this mean? I don't understand

melisaguevara (approved):
lgtm! 👍🏼

daywiss:
@amateima this has been updated for the upcoming changes to deposit ids and rebased against #163

packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
daywiss: not sure, i can remove this

james-a-morris: +1 on removal unless it's needed

packages/indexer-database/src/migrations/1735922359359-RelayHashInfo.ts
james-a-morris, line 1: Would it make sense to associate this table with the historic price through FK relationships?

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -154,6 +155,8 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
       timeToProcessAnciliaryEvents: finalPerfTime - timeToProcessDeposits,
       finalTime: finalPerfTime - startPerfTime,
     });
+    // publish new relays to workers to fill in prices
+    await this.publishNewRelays(storedEvents.fills);
```
james-a-morris, line 159: This should go above the timing and be included in timings since we're trying to track system time

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,280 @@
...
+
+export type PriceMessage = ss.Infer<typeof PriceMessage>;
+
+// Convert now to a consistent price timestamp yesterday for lookup purposes
+export function yesterday(now: Date) {
+  return DateTime.fromJSDate(now)
+    .minus({ days: 1 })
+    .set({ hour: 23, minute: 59, second: 0, millisecond: 0 })
+    .toJSDate();
+}
```
james-a-morris, line 27: Did we add a timing library to handle this?

packages/indexer/src/messaging/priceWorker.ts
```diff
@@ -0,0 +1,280 @@
...
+      where: {
+        date: priceTime,
+        baseCurrency,
+        quoteCurrency,
+      },
+    });
+    // we have this price at this time in the db
+    if (cachedPrice) {
+      return Number(cachedPrice.price);
+    }
```
james-a-morris, line 78: Where do we cache this? I don't see it being set later on

packages/indexer-database/src/migrations/1735922359359-RelayHashInfo.ts
daywiss: maybe but i dont think its needed. @amateima what do you think?

amateima (approved):

PR #144, opened 2025-01-02
fix: profile events storage relatively to insertion timestamp

amateima:
For more accurate profiling, we need to compute the time difference relatively to the time the event was inserted in the DB, not the current `now()` server timestamp at logging time.

In addition, now we profile fill events storing too

### DIFF
--- a/packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
+++ b/packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
@@ -114,7 +114,56 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
     );
     await this.updateNewDepositsWithIntegratorId(newInsertedDeposits);
     await this.spokePoolProcessor.process(storedEvents);
+    this.profileStoreEvents(storedEvents);
   }
+
+  /**
+   * Log the time that it took to store the events from the moment they were emitted onchain
+   * @param events
+   */
+  private profileStoreEvents(events: StoreEventsResult) {
+    const insertedDeposits = indexerDatabaseUtils.filterSaveQueryResults(
+      events.deposits,
+      SaveQueryResultType.Inserted,
+    );
+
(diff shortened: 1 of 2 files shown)


packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -114,7 +114,56 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
...
+      this.logger.debug({
+        at: "SpokePoolIndexerDataHandler#profileStoreEvents",
+        message: "V3FundsDeposited event profile",
+        depositId: event.depositId,
+        originChainId: event.originChainId,
+        timeDifference,
+        createdAt: event.createdAt,
+        blockTimestamp: event.blockTimestamp,
+      });
+    });
```
james-a-morris, line 144: OOC would it make sense to spot-check a few of these values? Since this is an O(N) operation, it could add noticeable milliseconds to the process, and it seems worth optimizing to shave down execution time as much as possible

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -114,7 +114,56 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
     );
     await this.updateNewDepositsWithIntegratorId(newInsertedDeposits);
     await this.spokePoolProcessor.process(storedEvents);
+    this.profileStoreEvents(storedEvents);
```
james-a-morris, line 117: Could the two async functions above this be called in parallel?

daywiss (approved):
nice changes

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
amateima: at some point we might reduce the number of logs, yes. I will introduce an ENV var to disable this while doing large backfills

PR #146, opened 2025-01-08
fix: fix building executed bundles

amateima:
**Problem**

In the `BundleIncludedEventsService`, I identified cases when the chain was included in the bundle, but the SpokePool contract for that chain was not deployed yet. We were trying to instantiate and update a SpokePool client with a block range at which the contract was not deployed yet. 

**Solution**
In this case, when building executed bundles, we skip the SpokePool client if the deployment block is > than the bundle end block

### DIFF
--- a/packages/indexer/src/services/BundleIncludedEventsService.ts
+++ b/packages/indexer/src/services/BundleIncludedEventsService.ts
@@ -1,7 +1,10 @@
 import * as across from "@across-protocol/sdk";
+import { getDeployedBlockNumber } from "@across-protocol/contracts";
 import Redis from "ioredis";
 import winston from "winston";
+
 import { DataSource, entities } from "@repo/indexer-database";
+
 import { BaseIndexer } from "../generics";
 import { BundleRepository } from "../database/BundleRepository";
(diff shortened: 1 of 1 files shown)


packages/indexer/src/services/BundleIncludedEventsService.ts
```diff
@@ -191,12 +194,37 @@ export class BundleIncludedEventsService extends BaseIndexer {
...
+        );
+        this.logger.debug({
+          at: "Indexer#BundleIncludedEventsService#getSpokeClientsForLookbackBlockRange",
+          message: `Instantiate SpokePool client for chain ${chainId}`,
+          deployedBlockNumber,
+          startBlock,
+          endBlockWithBuffer,
+        });
+        // A chain can be included in the bundle even if the SpokePool is not deployed yet
+        // In this case, the SpokePool client will not be instantiated and updated
```
james-a-morris, line 210: This isn't the case as we reference information about a spokepool in the bundle. The dataworker would crash + not produce a bundle if a spoke pool did not exist and a chain is included in `CHAIN_IDs`.

PR #156, opened 2025-01-22
fix: change depositId type for RelayHashInfo table

melisaguevara:


james-a-morris (approved):

amateima (approved):
Is this migration expected to take longer because of the table index?

daywiss (approved):

melisaguevara:
> Is this migration expected to take longer because of the table index?
Yes, I expect it to take longer than the previous ones

PR #157, opened 2025-01-22
fix: update repaymentChainId type and add new field for executedRelayerRefundRoot

melisaguevara:


packages/indexer-database/src/entities/evm/ExecutedRelayerRefundRoot.ts
```diff
@@ -37,6 +37,9 @@ export class ExecutedRelayerRefundRoot {
   @Column({ type: "jsonb" })
   refundAddresses: string[];
 
+  @Column({ nullable: true })
+  deferredRefunds: boolean;
+
```
melisaguevara, line 42: I'm not sure yet if this will be part of the SDK.

james-a-morris (approved):

amateima (approved):

daywiss (approved):

PR #162, opened 2025-01-23
improve(performance): disable quote block lookup in spoke indexer

james-a-morris:
This change removes the resolution of `quoteBlockNumber` from the spoke event indexer. It currently is not used in the context of deposit/fill lookups. This change does **not** remove the resolution of the `quoteBlockNumber` in the context of bundle creation/resolution.

Reason:
The `quoteBlockNumber` resolution in the SpokePoolClient requires a binary search to be completed from the deployment block. Since our calls are on increasing ranges and since we don't use the resolved value, we simply burn time resolving the values. 

Below is an example of the time it takes to resolve `v3FundsDeposited` which has this lookup. Note that it is 99% of the 120 seconds taken to update the spoke pool.

```
2025-01-23 16:22:15 [[34mdebug[39m]: {
  "at": "SpokePoolClient",
  "chainId": 1,
  "message": "SpokePool client for chain 1 updated!",
  "nextFirstBlockToSearch": 21079185,
  "elapsedMs": 122322.120625,
  "tokensBridgedTime": 0.10362499999973807,
  "v3FundsDepositedTime": 122198.531792,
  "requestedSpeedUpV3DepositTime": 0.22808299999451265,
  "requestedV3SlowFillTime": 1.4373750000086147,
  "filledV3RelayTime": 120.83100000000559,
  "enabledDepositRouteTime": 0.0027919999847654253,
  "relayedRootBundleTime": 0.07720800000242889,
  "executedRelayerRefundRootTime": 0.7112079999933485,
  "bot-identifier": "NO_BOT_ID",
  "run-identifier": "d2b7f368-ba0d-46aa-93a6-600b040dfdf3"
}
```

### DIFF
--- a/packages/indexer/src/redis/redisCache.ts
+++ b/packages/indexer/src/redis/redisCache.ts
@@ -8,7 +8,7 @@ export class RedisCache implements across.interfaces.CachingMechanismInterface {
     if (result === null) return result;
     return JSON.parse(result);
   }
-  set<ObjectType, OverrideType>(
+  set<ObjectType, _OverrideType>(
     key: string,
     value: ObjectType,
     ttl?: number,
--- a/packages/indexer/src/utils/clients/SpokePoolClient.ts
+++ b/packages/indexer/src/utils/clients/SpokePoolClient.ts
@@ -1,0 +1,42 @@
+import * as across from "@across-protocol/sdk";
+import { Contract } from "ethers";
+import winston from "winston";
+
+export class SpokePoolClient extends across.clients.SpokePoolClient {
+  constructor(
+    logger: winston.Logger,
+    spokePool: Contract,
+    hubPoolClient: across.clients.HubPoolClient | null,
+    chainId: number,
+    deploymentBlock: number,
+    eventSearchConfig?: across.utils.MakeOptional<
+      across.utils.EventSearchConfig,
+      "toBlock"
+    >,
+    private disableQuoteBlockLookup = false,
+  ) {
+    super(
+      logger,
+      spokePool,
+      hubPoolClient,
+      chainId,
+      deploymentBlock,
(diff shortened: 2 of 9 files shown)


packages/indexer/src/redis/redisCache.ts
```diff
@@ -8,7 +8,7 @@ export class RedisCache implements across.interfaces.CachingMechanismInterface {
     if (result === null) return result;
     return JSON.parse(result);
   }
-  set<ObjectType, OverrideType>(
+  set<ObjectType, _OverrideType>(
```
james-a-morris, line 11: QoL - we need to include the type but we should use `_` to specify that this field is unused.

packages/indexer/src/utils/clients/SpokePoolClient.ts
james-a-morris, line 1: To avoid needing any SDK change we can reference this custom behavior here

melisaguevara (commented):
This is a really good catch! 🙏🏼 
Looks good. Just one comment, I think we need to delete the field from the event before inserting in the database here https://github.com/across-protocol/indexer/blob/stage/packages/indexer/src/database/SpokePoolRepository.ts#L48

james-a-morris:
> This is a really good catch! 🙏🏼 Looks good. Just one comment, I think we need to delete the field from the event before inserting in the database here https://github.com/across-protocol/indexer/blob/stage/packages/indexer/src/database/SpokePoolRepository.ts#L48

Good callout! This wasn't throwing errors in my local testing but I've added it to be sure: https://github.com/across-protocol/indexer/pull/162/commits/910c90ae91f5c3244be7e33fe8d735d86aa53365

melisaguevara (approved):

PR #163, opened 2025-01-23
feat: support duplicated deposits and handle fill event matching

melisaguevara:
Closes [ACX-3663](https://linear.app/uma/issue/ACX-3663/indexer-update-unique-constraints) and [ACX-3665](https://linear.app/uma/issue/ACX-3665/indexer-ensure-fill-events-are-matched-with-a-single-deposit)

This PR addresses some of the changes needed to handle duplicated deposits in the Indexer.

1. Adjust unique constraints:

    - in v3FundsDeposited:  changing from `depositId` + `originChainId` to `relayHash` + `blockNumber` + `logIndex`.
    - in RelayHashInfo: changing from relayHash to `relayHash` + `depositEventId`. We use `depositEventId` since deposit instances are uniquely identified in the `v3FundsDeposited` table.

2. Adjust event matching logic in relayHashInfo table:

Previously, the uniqueness of relayHash allowed us to use UPSERT queries when inserting new events. However, since relayHash is no longer unique and the relayHashInfo table must now handle all potential edge cases caused by colliding deposits, finding a single unique constraint to enable UPSERT queries was not feasible.

To address this, we are transitioning to a SELECT + INSERT/UPDATE approach. The downside of this approach is the potential for race conditions when multiple processes are handling events with the same relayHash. To prevent those, we'll be using [advisory locks](https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS), which are application-controlled locks that allow us to synchronize access to specific resources without locking the entire table.
They work by acquiring an exclusive lock on a numeric key, preventing other transactions from modifying the same resource until the lock is released.

For our use case, we will use (originChainId, 32-bit integer representation of relayHash) as the lock key, ensuring that concurrent processes do not interfere with each other when handling events with the same relayHash.

### DIFF
--- a/packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
+++ b/packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
@@ -7,9 +7,10 @@ import {
 } from "typeorm";
 
 @Entity({ schema: "evm" })
-@Unique("UK_v3FundsDeposited_depositId_originChainId", [
-  "depositId",
-  "originChainId",
+@Unique("UK_v3FundsDeposited_relayHash_block_logIdx", [
+  "relayHash",
+  "blockNumber",
+  "logIndex",
 ])
 export class V3FundsDeposited {
   @PrimaryGeneratedColumn()
--- a/packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
+++ b/packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
@@ -20,7 +20,7 @@ export class V3FundsDeposited {
   relayHash: string;
 
   @Column({ type: "decimal" })
-  depositId: number;
+  depositId: string;
 
   @Column()
   originChainId: number;
--- a/packages/indexer/src/services/spokePoolProcessor.ts
+++ b/packages/indexer/src/services/spokePoolProcessor.ts
@@ -207,11 +207,25 @@ export class SpokePoolProcessor {
 
     const upsertResults = [];
     for (const item of data) {
-      upsertResults.push(
-        await relayHashInfoRepository.upsert(item, ["relayHash"]),
-      );
+      // In case of colliding deposits, match the fill only with the first deposit event we stored in the database
+      const existingRow = await relayHashInfoRepository.findOne({
+        where: { relayHash: item.relayHash },
+        order: { depositEventId: "ASC" },
+      });
+      if (!existingRow) {
+        const insertedRow = await relayHashInfoRepository.insert(item);
+        upsertResults.push(insertedRow);
+      } else {
+        const updatedRow = await relayHashInfoRepository.update(
+          {
+            relayHash: item.relayHash,
+            depositEventId: existingRow.depositEventId,
+          },
+          item,
+        );
+        upsertResults.push(updatedRow);
+      }
     }
-
     this.logger.debug({
       at: "Indexer#SpokePoolProcessor#assignSpokeEventsToRelayHashInfo",
       message: `${eventType} events associated with RelayHashInfo`,
--- a/packages/indexer/src/services/spokePoolProcessor.ts
+++ b/packages/indexer/src/services/spokePoolProcessor.ts
@@ -200,20 +200,18 @@ export class SpokePoolProcessor {
         };
 
         // Start a transaction
-        const queryRunner = this.postgres.createQueryRunner();
-        await queryRunner.connect();
-        await queryRunner.startTransaction();
-        const relayHashInfoRepository = queryRunner.manager.getRepository(
-          entities.RelayHashInfo,
-        );
-        try {
+        await this.postgres.transaction(async (transactionalEntityManager) => {
+          const relayHashInfoRepository =
+            transactionalEntityManager.getRepository(entities.RelayHashInfo);
+
           // Convert relayHash into a 32-bit integer for database lock usage
           const lockKey = this.relayHashToInt32(item.relayHash);
           // Acquire a lock to prevent concurrent modifications on the same relayHash.
-          await queryRunner.query(`SELECT pg_advisory_xact_lock($2, $1)`, [
-            item.originChainId,
-            lockKey,
-          ]);
+          // The lock is automatically released when the transaction commits or rolls back.
+          await transactionalEntityManager.query(
+            `SELECT pg_advisory_xact_lock($2, $1)`,
+            [item.originChainId, lockKey],
+          );
 
           // Retrieve an existing entry that either:
           // - Matches the relayHash and has no associated depositEventId.
@@ -241,14 +239,7 @@ export class SpokePoolProcessor {
             );
             updateResults.push(updatedRow);
           }
-          await queryRunner.commitTransaction();
-        } catch (error) {
-          await queryRunner.rollbackTransaction();
-          throw error;
-        } finally {
-          // Release transaction resources; locks acquired will be automatically released.
-          await queryRunner.release();
-        }
+        });
       }),
     );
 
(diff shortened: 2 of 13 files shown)


james-a-morris (commented):
This is marked as draft - is this ready for review?

packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
```diff
@@ -7,9 +7,10 @@ import {
 } from "typeorm";
 
 @Entity({ schema: "evm" })
-@Unique("UK_v3FundsDeposited_depositId_originChainId", [
-  "depositId",
-  "originChainId",
+@Unique("UK_v3FundsDeposited_relayHash_block_logIdx", [
+  "relayHash",
+  "blockNumber",
+  "logIndex",
```
daywiss, line 13: @amateima and i discussed using [chainid, blockhash, txhash, logindex] for uniquely identifying events. this still on the table?

melisaguevara: We can def do this but I would handle it in a separate PR.

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -154,9 +165,9 @@ export class SpokePoolProcessor {
     const eventTypeToField = {
       [SpokePoolEvents.V3FundsDeposited]: "depositEventId",
       [SpokePoolEvents.FilledV3Relay]: "fillEventId",
-      [SpokePoolEvents.RequestedV3SlowFill]: "requestSlowFillEventId",
+      [SpokePoolEvents.RequestedV3SlowFill]: "slowFillRequestEventId",
```
melisaguevara, line 168: While testing I found this field had a wrong name 😬

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -178,20 +189,56 @@ export class SpokePoolProcessor {
       };
     });
```
amateima, line 191: Personally I feel that `assignSpokeEventsToRelayHashInfo` starts to cover too many cases. The multiple optional fields, `if` instructions and `&&` operators can have a negative impact during debugging for first time code readers. What if we would implement separate functions `assignDepositEventsToRelayHashInfo`, `assignFillEventsToRelayHashInfo`, etc. I know that they would have some duplicated code, but I think there are some nuances in handling these events that need to be highlighted

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -178,20 +189,56 @@ export class SpokePoolProcessor {
...
+      } else {
+        const updatedRow = await relayHashInfoRepository.update(
+          {
+            id: existingRow.id,
+            relayHash: item.relayHash,
+          },
+          item,
+        );
+        updateResults.push(updatedRow);
+      }
```
amateima, line 231: The problem with separate `SELECT` and `UPDATE` / `INSERT` queries is that the DB state can change between the two operations. Is it possible to replace it with `UPSERT` queries?

amateima:
Just to confirm, this PR doesn't cover pre-fills, right?

packages/indexer/src/services/spokePoolProcessor.ts
melisaguevara: I think that by replacing `assignSpokeEventsToRelayHashInfo` with one function for each event this should be possible. I'll give it a try.

melisaguevara:
> Just to confirm, this PR doesn't cover pre-fills, right?

@amateima My understanding is that pre-fills are just fills that come before a deposit. If that's the case then I think we are covered as we already store fills in the database independently of if there's a deposit or not.
But is there anything else we have to take into account that I might be missing?

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -150,75 +144,303 @@ export class SpokePoolProcessor {
...
+      totalTime:
+        timeToUpdateRefundedDepositsEnd -
+        timeToAssignSpokeEventsToRelayHashInfoStart,
     });
   }
 
   /**
    * Updates relayHashInfo table to include recently stored events
-   * @param events An array of already stored deposits, fills or slow fill requests
+   * @param events An object of already stored deposits, fills and slow fill requests
```
melisaguevara, line 162:
```suggestion
   * @param events An object with stored deposits, fills and slow fill requests
```

daywiss (approved):
looks good. good find with the locking

packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
james-a-morris: +1 on a separate PR

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -351,4 +573,31 @@ export class SpokePoolProcessor {
...
+      // This shifts the hash left by 5 bits, subtracts itself, and adds the character code
+      hash = (hash << 5) - hash + chr;
+
+      // Convert the result into a 32-bit integer by forcing it into the signed integer range
+      hash |= 0;
+    }
+
+    // Return the final computed 32-bit integer hash
+    return hash;
+  }
```
james-a-morris, line 602: Is this something that's not defined in the SDK or the relayer? It could be a good candidate to reduce multiple definitions

james-a-morris (commented):
few more comments

packages/indexer/src/services/spokePoolProcessor.ts
melisaguevara: I don't think so. I haven't heard of any other use case where we need to re-hash the relayHash.

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -178,20 +189,56 @@ export class SpokePoolProcessor {
...
+        // 3. A matching relayHash exists without an associated depositEventId.
+        queryBuilder.andWhere(
+          `"depositEventId" IS NULL OR "depositEventId" = :itemEventId`,
+          { itemEventId: item.depositEventId },
+        );
+      }
+
+      const existingRow = await queryBuilder
+        .orderBy(`"depositEventId"`, "ASC") // Get only the first inserted deposit if it exists
+        .getOne();
```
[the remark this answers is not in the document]
melisaguevara: `depositEventId` has a unique constraint, so I think it's already indexed.

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -150,75 +144,303 @@ export class SpokePoolProcessor {
...
+          fillDeadline: event.fillDeadline,
+          depositEventId: event.id,
           depositTxHash: event.transactionHash,
-        }),
-        ...(eventType === SpokePoolEvents.FilledV3Relay && {
-          status: RelayStatus.Filled,
+        };
+
+        // Start a transaction
+        const queryRunner = this.postgres.createQueryRunner();
```
amateima, line 203: I think we would prefer opening transactions on the dataSource directly, instead of opening a new connection

```javascript
await dataSource.transaction(async (transactionalEntityManager) => {
    // execute queries using transactionalEntityManager
})
```

if we use this, than we don't have to commit or rollback manually

amateima (changes requested):

packages/indexer/src/services/spokePoolProcessor.ts
melisaguevara: Changed here [643104e](https://github.com/across-protocol/indexer/pull/163/commits/643104eabb45678566f18b70d992f78d57e4858a)

packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -38,13 +37,44 @@ export class SpokePoolProcessor {
       SaveQueryResultType.Updated,
     );
```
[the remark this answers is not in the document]
melisaguevara: Removed here [f20704f](https://github.com/across-protocol/indexer/pull/163/commits/f20704fa7f97d2253401e444dcd3e702bd87bb4d)

PR #166, opened 2025-01-28
feat: handle depositId bigInt type from sdk

melisaguevara:
- Updating SDK to new version where depositId went from number to BigInt
- Adapting our code to this new type. The main change is that we'll be handling depositIds as a string in TypeORM entities.

### DIFF
--- a/packages/indexer-database/package.json
+++ b/packages/indexer-database/package.json
@@ -25,7 +25,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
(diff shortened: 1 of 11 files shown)


james-a-morris (commented):
Does this account for all repos that import the sdk?

melisaguevara:
> Does this account for all repos that import the sdk?

Yes. We use the sdk both in indexer and indexer-database packages and this PR covers the changes needed in both.

amateima (approved):

james-a-morris (approved):

PR #171, opened 2025-02-05
feat: handle unfinalised deposit events

melisaguevara:
With our recent update to how reorgs are handled for deposits (see [this change](https://github.com/across-protocol/indexer/pull/163/files#diff-cea1547b0947849dc19bac1c8b54d7c40a0f5975f4cba996111e8d560eab5db0L66-R68)), unfinalized events will now be stored in the `deposits` table. This PR introduces a new step in the `spokePoolIndexerDataHandler` to soft delete these events. The deleted events are then sent to the processor, where we'll update the related `relayHashInfo` records.

While updating `relayHashInfo`, we need to ensure that relations to other events (fills, slow fill requests, refunds) are not lost. The logic for handling this is as follows:

1. For each deleted deposit, retrieve the related `relayHashInfo` row. Since this is a one-to-one relation, we can be sure only one related row exists.
2. Check if other events are already associated with that `relayHashInfo` record:
   - If no other events exist, the row can be safely deleted.
   - If other events are already related, check for additional rows with the same `relayHash`:
     - If additional rows exist, transfer the deposit relation from one of them to the row where the other events are stored, then delete the original row.
     - If no other rows exist, set the `depositEventId` and `depositTxHash` fields to `undefined` instead of deleting the row.

 A visual example for clarity:
 
image

### DIFF
--- a/packages/indexer-database/src/utils/BlockchainEventRepository.ts
+++ b/packages/indexer-database/src/utils/BlockchainEventRepository.ts
@@ -109,4 +109,43 @@ export class BlockchainEventRepository {
       result: SaveQueryResultType.Nothing,
     };
   }
+
+  protected async deleteUnfinalisedEvents<Entity extends ObjectLiteral>(
+    chainId: number,
+    chainIdColumnIdentifier: string,
+    lastFinalisedBlock: number,
+    entity: EntityTarget<Entity>,
+  ): Promise<Entity[]> {
+    const entityMetadata = this.postgres.getMetadata(entity);
+    const columns = entityMetadata.columns.map((column) => column.propertyName);
+    const hasChainIdTargetColumn = columns.includes(chainIdColumnIdentifier);
+    const hasDeletedAtColumn = columns.includes("deletedAt");
+
+    if (
+      entityMetadata.schema !== "evm" ||
+      !hasChainIdTargetColumn ||
+      !hasDeletedAtColumn
+    ) {
+      this.logger.error({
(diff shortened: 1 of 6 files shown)


packages/indexer-database/src/utils/BlockchainEventRepository.ts
```diff
@@ -109,4 +109,43 @@ export class BlockchainEventRepository {
...
+        hasChainIdTargetColumn,
+        hasDeletedAtColumn,
+      });
+      throw new Error(`Cannot delete events of ${entityMetadata.name} entity`);
+    }
+
+    const repository = this.postgres.getRepository(entity);
+    const deletedRows = await repository
+      .createQueryBuilder()
+      .softDelete()
```
daywiss, line 142: what is soft delete?

amateima (approved):

packages/indexer-database/src/utils/BlockchainEventRepository.ts
melisaguevara: Soft delete doesn't delete the row, it just sets a value for 'deletedAt' column. And then any other queries made to the table are made including an extra filter `WHERE deletedAt IS NOT NULL`

melisaguevara:
> very delicate update, lgtm. what can we do if something is wrong with the logic here once it makes these updates?

Yes, I agree. We will have the deposit events in the raw events table so we should be able to debug and restore the changes in relayHashInfo if needed.

PR #173, opened 2025-02-06
feat: support doctor who

melisaguevara:
+ Support Doctor Who
+ Rename Cher constants to Soneium


james-a-morris (changes requested):
looks good but ideally these two changes should be separate PRs (change soneium, add dr who)

melisaguevara:
> looks good but ideally these two changes should be separate PRs (change soneium, add dr who)

They can't. The constants package version doesn't include Cher anymore.

james-a-morris (approved):

PR #178, opened 2025-02-07
fix: add internal relay hash

amateima:


melisaguevara (approved):

packages/indexer-database/src/migrations/1738886982012-RelayHashInfo.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class RelayHashInfo1738886982012 implements MigrationInterface {
+  name = "RelayHashInfo1738886982012";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "relay_hash_info" ADD "internalHash" character varying`,
+    );
+    await queryRunner.query(
+      `update "relay_hash_info" set "internalHash" = "relayHash" where "relayHash" is not null;`,
```
mrice32, line 11: So basically, elements that have already been ingested will have their relayhash used as the internal hash. Future ingestion will use the new internalHash construction. Is that right?

amateima: Yes, that's the plan

PR #181, opened 2025-02-08
fix: use internal hash for bundle events

melisaguevara:


packages/indexer/src/database/BundleRepository.ts
```diff
@@ -593,7 +593,11 @@ export class BundleRepository extends utils.BaseRepository {
         events.map((event) => {
           return {
             bundleId,
-            relayHash: across.utils.getRelayHashFromEvent(event),
+            relayHash: getInternalHash(
+              event,
+              event.messageHash,
+              event.destinationChainId,
+            ),
```
melisaguevara, line 600: We have two different functions for formatting bundle events, one for fills and one for all the other events. We changed this for fills but not for the remaining events.

daywiss (approved):

amateima (approved):

PR #183, opened 2025-02-11
fix: refunded deposits webhook

melisaguevara:


packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -686,7 +686,7 @@ export class SpokePoolProcessor {
             })
             .returning("*")
             .execute();
-          updatedRows.push(updatedRow.raw);
+          updatedRows.push(updatedRow.raw[0]);
```
melisaguevara, line 689: updatedRows was storing a nested array of entities instead of an array of entities.

amateima (approved):

PR #185, opened 2025-02-11
improve(db): load block range from postgres

james-a-morris:


packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -133,6 +128,31 @@ export class Indexer {
     );
   }
 
+  private async loadProgressFromDatabase(): Promise<BlockRangeResult> {
+    const indexerProgressInfo = await this.dataSource
+      .getRepository(entities.IndexerProgressInfo)
+      .findOne({
+        where: {
+          id: this.dataHandler.getDataIdentifier(),
```
daywiss, line 136: is it assumed only a single process is reading/writing to this, or does that matter?

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -57,7 +57,7 @@ export class Indexer {
         // if the previous block range was processed successfully or if this is the first loop iteration,
         // get the next block range to process
         if (blockRangeProcessedSuccessfully || !blockRangeResult) {
-          blockRangeResult = await this.getBlockRange();
+          blockRangeResult = await this.loadProgressFromDatabase();
```
daywiss, line 60: it looks like progress was already being stored to postgres, so this is like a continuation of that work to start loading that data?

daywiss (approved):
looks good just some questions!

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -133,6 +128,31 @@ export class Indexer {
...
+  private async loadProgressFromDatabase(): Promise<BlockRangeResult> {
+    const indexerProgressInfo = await this.dataSource
+      .getRepository(entities.IndexerProgressInfo)
+      .findOne({
+        where: {
+          id: this.dataHandler.getDataIdentifier(),
+        },
+      });
+    if (across.utils.isDefined(indexerProgressInfo)) {
+      const fromBlock = indexerProgressInfo.lastFinalisedBlock;
```
melisaguevara, line 140: We were increasing `lastFinalisedBlock` by one before. Is that not needed anymore?

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -141,24 +161,10 @@ export class Indexer {
    *  i.e no new blocks have been mined, then the block range is `undefined`.
    */
   private async getBlockRange(): Promise<BlockRangeResult> {
```
melisaguevara, line 163: Wdyt about moving the call to `loadProgressFromDatabase()` inside this function?
I think it is slightly easier to read compared to `loadProgressFromDatabase` calling `getBlockRange`

packages/indexer/src/data-indexing/service/Indexer.ts
amateima: > is it assumed only a single process is reading/writing to this

yes

packages/indexer/src/data-indexing/service/Indexer.ts
amateima: I am aligned with Melisa on this. I think would be more intuitive to call `loadProgressFromDatabase()` here

packages/indexer/src/data-indexing/service/Indexer.ts
james-a-morris: https://github.com/across-protocol/indexer/pull/185/commits/7ffad2618f46643e49dbe078d37d7e2dff8847d4

packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -141,24 +161,10 @@ export class Indexer {
...
       latestBlockNumber - this.config.finalisedBlockBufferDistance;
-
-    if (lastBlockFinalisedStored === lastFinalisedBlockOnChain) {
-      return {
-        latestBlockNumber,
-        blockRange: undefined,
-        lastFinalisedBlock: lastFinalisedBlockOnChain,
-        isBackfilling: false,
-      };
-    }
```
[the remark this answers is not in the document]
james-a-morris: Fair call - our frequency will potentially cause overlapping runs. I.e. we should be able to query mainnet more than once every 13 seconds which would cause this edge case to be hit.

https://github.com/across-protocol/indexer/pull/185/commits/2ab0b60469060c263a58421ba4ab85be3a2ed29b

packages/indexer/src/data-indexing/service/Indexer.ts
james-a-morris: I would prefer to keep these calls separate. `getBlockRange()` specifically calls into RPC providers where `loadProgressFromDatabase()` specifically calls into the database (and falls back to the `getBlockRange()`). They are encapsulated and scoped

james-a-morris:
@amateima @melisaguevara I've updated the conditional statement + rearranged function scope here: https://github.com/across-protocol/indexer/pull/185/commits/a2c817ec71f4111548d26cc4576fa15db12faf27

melisaguevara (commented):
looks good! Just one comment

melisaguevara (approved):

melisaguevara (approved):

amateima (approved):

PR #196, opened 2025-02-19
feat: add swap events

amateima:


packages/indexer/src/services/spokePoolProcessor.ts
```diff
@@ -732,4 +737,25 @@ export class SpokePoolProcessor {
...
+    );
+    await Promise.all(
+      depositSwapPairs.map((depositSwapPair) =>
+        relayHashInfoRepository.update(
+          { depositEventId: depositSwapPair.deposit.id },
+          {
+            swapBeforeBridgeEventId: depositSwapPair.swapBeforeBridge.id,
+          },
+        ),
+      ),
```
james-a-morris, line 758: Can we do this in one call?

packages/indexer/src/web3/EventDecoder.ts
```diff
@@ -0,0 +1,48 @@
+import { ethers } from "ethers";
+import { SwapBeforeBridgeEvent } from "./model/events";
+import { SwapAndBridgeBase__factory } from "@across-protocol/contracts";
+
+export class EventDecoder {
+  static decodeSwapBeforeBridgeEvents(
+    receipt: ethers.providers.TransactionReceipt,
+  ) {
+    const swapBeforeBridgeEventTopic =
+      "0x646284e396b68ff4b4f34e0aa97bcdb9c100f5b44a20da5c475f627039853841";
```
james-a-morris, line 10: Should we hardcode this elsewhere or pull this in from constants/contracts/sdk?

packages/indexer/src/services/spokePoolProcessor.ts
amateima: One call is fine, I preferred to do it this more granular way because of the issues we had when doing queries in batches (conflicts, max rows that can be updated in a single query, etc)

packages/indexer/src/web3/EventDecoder.ts
amateima: This function is meant to encapsulate all the technical details required to parse swap events, including constants. Not sure if this can be pulled from other package

PR #199, opened 2025-02-19
fix(bundle-service): remove unneeded error condition

james-a-morris:
This error condition rarely appears as it will only happen when a database is fresh and an unlucky race condition occurs where the `BundleEventsProcessor` is fired before enough data is available. In all other `assign{}ToBundle` functions we ignore this case, but in the `assignExecutionsToBundle` we throw an error. I believe this is unwarranted as we gracefully ignore and re-process the failing event on the next loop until enough data is available.

### DIFF
--- a/packages/indexer/src/services/bundles.ts
+++ b/packages/indexer/src/services/bundles.ts
@@ -197,14 +197,7 @@ async function assignExecutionsToBundle(
             logIndex,
           );
         if (!proposedBundle) {
-          logger.error({
-            at: "Indexer#BundleEventsProcessor#assignExecutionsToBundle",
-            message: "Unable to find a proposed bundle for the given execution",
-            executionId: id,
-          });
-          throw new Error(
-            `Unable to find a proposed bundle for the given execution ${id}`,
-          );
+          return undefined;
         }
         return {
           bundleId: proposedBundle.bundle.id,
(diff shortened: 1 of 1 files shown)


melisaguevara:
> I believe this is unwarranted as we gracefully ignore and re-process the failing event on the next loop until enough data is available.

Why do you think more data will be available in the next runs?
I might be missing something, but since the order of events is always `proposedRootBundle` first and `rootBundleExecuted` next, it seems like this error is flagging that we missed a proposal.

james-a-morris:
> > I believe this is unwarranted as we gracefully ignore and re-process the failing event on the next loop until enough data is available.
> 
> Why do you think more data will be available in the next runs? I might be missing something, but since the order of events is always `proposedRootBundle` first and `rootBundleExecuted` next, it seems like this error is flagging that we missed a proposal.

The unfortunate race condition here on the initial database load is: 
1. SpokePoolIndexer runs first causing data to be populated about a bundle event
2. The BundleProcessor runs and fails
3. The HubPoolIndexer runs and loads its bundle data
4. (the system resolves because the BundleProcessor is out of band looking for events that haven't been assigned)

melisaguevara:
> The unfortunate race condition here on the initial database load is:
> 
> 1. SpokePoolIndexer runs first causing data to be populated about a bundle event
> 2. The BundleProcessor runs and fails
> 3. The HubPoolIndexer runs and stores its proposed root bundle data to the database
> 4. The BundleProcessor runs again and succeeds
>    (the system resolves because the BundleProcessor is out of band looking for events that haven't been assigned)

Which data are you referring to in step 1?
Does the BundleProcessor use any event from the SpokePools? I think they're all from the HubPool

james-a-morris:
Closing in light of #201

PR #201, opened 2025-02-21
feat: refactor bundle service in hub indexer

james-a-morris:
WIP: need to test

### DIFF
--- a/packages/indexer/src/database/HubPoolRepository.ts
+++ b/packages/indexer/src/database/HubPoolRepository.ts
@@ -27,16 +27,18 @@ export class HubPoolRepository extends utils.BaseRepository {
     });
 
     const chunkedEvents = across.utils.chunk(formattedEvents, this.chunkSize);
-    await Promise.all(
+    const chunkedResults = await Promise.all(
       chunkedEvents.map((eventsChunk) =>
         this.postgres
-          .createQueryBuilder(entities.ProposedRootBundle, "b")
-          .insert()
-          .values(eventsChunk)
-          .orUpdate(["finalised"], ["transactionHash"])
-          .execute(),
+          .getRepository(entities.ProposedRootBundle)
+          .save(eventsChunk, {
+            chunk: this.chunkSize,
+            reload: true,
+          }),
       ),
     );
+
+    return chunkedResults.flat();
   }
 
   public async formatAndSaveRootBundleDisputedEvents(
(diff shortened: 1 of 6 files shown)


daywiss (approved):

amateima (commented):
I am investigating an error I get while testing it locally 🤔 
```
{
  "length":308,
  "name":"error",
  "severity":"ERROR",
  "code":"23505",
  "detail":"Key (\"transactionHash\")=(0xc694a80e0bec631e737ea7024bf8b1fc4b7d36014f424ccef0bd1a5a5a061ea3) already exists.",
  "schema":"evm",
  "table":"proposed_root_bundle",
  "constraint":"UK_proposedRootBundle_txHash",
  "file":"nbtinsert.c",
  "line":"649",
  "routine":"_bt_check_unique"
}
```

In the meantime, can we remove all references to `enableBundleEventsProcessor` and `ENABLE_BUNDLE_EVENTS_PROCESSOR`?

packages/indexer/src/database/HubPoolRepository.ts
```diff
@@ -25,18 +23,15 @@ export class HubPoolRepository extends utils.BaseRepository {
...
-          .values(eventsChunk)
-          .orUpdate(["finalised"], ["transactionHash"])
-          .execute(),
-      ),
-    );
+    const savedEvents =
+      await this.saveAndHandleFinalisationBatch<entities.ProposedRootBundle>(
+        entities.ProposedRootBundle,
+        formattedEvents,
+        ["transactionHash"],
```
melisaguevara, line 30: Just curious about your thoughts on start using blockNumber and logIndex as the standard unique constraints.

packages/indexer/src/database/HubPoolRepository.ts
```diff
@@ -104,20 +94,15 @@ export class HubPoolRepository extends utils.BaseRepository {
...
-          .execute(),
-      ),
-    );
+    const savedEvents =
+      await this.saveAndHandleFinalisationBatch<entities.RootBundleExecuted>(
+        entities.RootBundleExecuted,
+        formattedEvents,
+        ["chainId", "leafId", "groupIndex", "transactionHash"],
+        [],
+      );
```
melisaguevara, line 103: Is splitting the list needed here? We should be ok when indexing recent events, but could it be required for backfilling?

melisaguevara (commented):
looks good! Just 2 questions

packages/indexer/src/database/HubPoolRepository.ts
amateima: I think we are good because the number of hubpool events is not that big

amateima (approved):

PR #209, opened 2025-02-26
feat: test accurate handling of deleted deposits

melisaguevara:
This PR adds tests to ensure that RelayHashInfo correctly handles deleted and unfinalized deposits.

Test cases:
- Deposit - Unfinalized and no other events with same internalHash: If a deposit is not finalized and has no related events, its RelayHashInfo row is deleted.
- Deposit - Fill - Unfinalized Deposit: If a deposit is not finalized, but a fill exists, the deposit reference is removed, but the RelayHashInfo row remains to keep fill data.
- Deposit - Fill - Replacing Deposit - Unfinalized First Deposit: When a deposit is replaced, and the original is unfinalized, the deposit reference is transferred to the existing RelayHashInfo row that holds fill data and the redundant row is deleted.


packages/indexer-database/src/utils/BlockchainEventRepository.ts
```diff
@@ -16,7 +16,7 @@ export function filterSaveQueryResults<Entity extends ObjectLiteral>(
 export class BlockchainEventRepository {
   constructor(
     protected postgres: DataSource,
-    protected logger: winston.Logger,
+    protected logger?: winston.Logger,
```
james-a-morris, line 19: Did we want to make this optional? I'd imagine this is something we should require

packages/indexer/src/database/SpokePoolRepository.ts
```diff
@@ -6,7 +6,7 @@ import * as utils from "../utils";
 export class SpokePoolRepository extends dbUtils.BlockchainEventRepository {
   constructor(
     postgres: DataSource,
-    logger: winston.Logger,
+    logger?: winston.Logger,
```
james-a-morris, line 9: Similar to above

james-a-morris (commented):
Left a few questions - could be worth adding a winston transport that doesn't output anything to make sure we're not making the logger optional

daywiss (approved):

packages/indexer-database/src/utils/BlockchainEventRepository.ts
melisaguevara: Adressed in here: [9751c86](https://github.com/across-protocol/indexer/pull/209/commits/9751c8652e2ded916f1d9e6ffb53ebaa58042668)

PR #217, opened 2025-03-17
feat(svm): svm indexer data handler and solana constants

melisaguevara:
This PR:
- Adds Solana's constants.
- Adds a new `SvmSpokePoolIndexerDataHandler ` class that instantiates the svm spoke client to fetch and return events. This class will be used in other upcoming PRs.

### DIFF
--- a/packages/indexer/src/data-indexing/service/constants.ts
+++ b/packages/indexer/src/data-indexing/service/constants.ts
@@ -2,6 +2,7 @@ import { CHAIN_IDs } from "@across-protocol/constants";
 
 // taken from https://github.com/UMAprotocol/bot-configs/blob/ed878f5f80509ad4ca55c8200e40670ba50e3b26/serverless-bots/across-v2-bot-config.json#L330C1-L342C25
 const finalisedBlockBufferDistances: Record<number, number> = {
+  // Mainnets
   [CHAIN_IDs.ALEPH_ZERO]: 80,
   [CHAIN_IDs.ARBITRUM]: 240,
   [CHAIN_IDs.BASE]: 60,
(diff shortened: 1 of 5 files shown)


packages/indexer/src/redis/redisCache.ts
```diff
@@ -13,7 +13,7 @@ export class RedisCache implements across.interfaces.CachingMechanismInterface {
     value: ObjectType,
     ttl?: number,
   ): Promise<string | undefined> {
-    if (ttl !== undefined) {
+    if (ttl !== undefined && ttl !== Number.POSITIVE_INFINITY) {
```
melisaguevara, line 16: This is needed because our providers use Number.POSITIVE_INFINITY to specify no ttl should be set. See for example the relayer implementation [here](https://github.com/across-protocol/relayer/blob/master/src/utils/RedisUtils.ts#L44).

packages/indexer/src/data-indexing/service/constants.ts
```diff
@@ -15,13 +16,16 @@ const finalisedBlockBufferDistances: Record<number, number> = {
   [CHAIN_IDs.POLYGON]: 128,
   [CHAIN_IDs.REDSTONE]: 60,
   [CHAIN_IDs.SCROLL]: 40,
+  [CHAIN_IDs.SOLANA]: 40,
```
melisaguevara, line 19: Based on the [docs](https://solana.com/developers/guides/advanced/confirmation#fetch-blockhashes-with-the-appropriate-commitment-level), which state:

> Using the [default commitment](https://solana.com/docs/rpc#default-commitment) level finalized will eliminate any risk that the blockhash you choose will belong to a dropped fork. The tradeoff is that there is typically at least a 32 slot difference between the most recent confirmed block and the most recent finalized block.

amateima (approved):

daywiss (approved):

PR #218, opened 2025-03-17
feat(providersFactory): support svm providers

melisaguevara:
This PR modifies the `retryProvidersFactory` so it is able to instantiate both EVM and SVM providers based on the 'family' of the provided chain. For SVM, it will instantiate cached providers.

### DIFF
--- a/packages/indexer/src/data-indexing/service/AcrossIndexerManager.ts
+++ b/packages/indexer/src/data-indexing/service/AcrossIndexerManager.ts
@@ -83,7 +83,9 @@ export class AcrossIndexerManager {
         ),
       },
       hubPoolIndexerDataHandler,
-      this.retryProvidersFactory.getProviderForChainId(this.config.hubChainId),
+      this.retryProvidersFactory.getProviderForChainId(
+        this.config.hubChainId,
+      ) as across.providers.RetryProvider,
       this.logger,
       this.postgres,
     );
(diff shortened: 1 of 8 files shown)


packages/indexer/src/data-indexing/service/AcrossIndexerManager.ts
```diff
@@ -83,7 +83,9 @@ export class AcrossIndexerManager {
         ),
       },
       hubPoolIndexerDataHandler,
-      this.retryProvidersFactory.getProviderForChainId(this.config.hubChainId),
+      this.retryProvidersFactory.getProviderForChainId(
+        this.config.hubChainId,
+      ) as across.providers.RetryProvider,
```
melisaguevara, line 88: One drawback of this change is that we'll need to specify the expected type from the factory when we call `getProviderForChainId` as it could now return either an EVM or an SVM provider.

packages/indexer/src/utils/contractUtils.ts
```diff
@@ -176,42 +176,4 @@ export function getHubPoolClient(
...
-  pctRpcCallsLogged: number;
-  standardTtlBlockDistance: number;
-  noTtlBlockDistance: number;
-  providerCacheTtl: number;
-  nodeQuorumThreshold: number;
-  retries: number;
-  delay: number;
-  providerConfigs: [providerUrls: string, chainId: number][];
-  chainId: number;
-};
```
melisaguevara, line 191: I moved this to parseEnv.ts

packages/indexer/src/data-indexing/service/AcrossIndexerManager.ts
daywiss: typically you would have to provide another function like getSolanaProvider and getEthProvider to know the right type

daywiss (approved):

packages/indexer/src/data-indexing/service/AcrossIndexerManager.ts
melisaguevara: good call! I'll add those in a follow-up PR 👍🏼

PR #219, opened 2025-03-17
feat(svm): initial svm indexing version

melisaguevara:
This PR adds support for svm indexing.

- It adds two new classes, `EvmIndexer` and `SvmIndexer`, which extend the base `Indexer` class. These new classes are needed as the function to get the lates block/slot is different for each family.
- `AcrossIndexerManager` now instantiates either the EvmIndexer or the SvmIndexer based on the family of the chain we want to index. To enable this, the `SPOKEPOOL_CHAINS_ENABLED` list has been split into separate lists for EVM and SVM chains.

To test this locally with solana devnet you can checkout to this branch and include in your .env file:
```
SPOKEPOOL_CHAINS_ENABLED=133268194659241,...
RPC_PROVIDER_URLS_133268194659241=https://api.devnet.solana.com
```

### DIFF
--- a/packages/indexer/src/data-indexing/service/SvmSpokePoolIndexerDataHandler.ts
+++ b/packages/indexer/src/data-indexing/service/SvmSpokePoolIndexerDataHandler.ts
@@ -53,10 +53,7 @@ export class SvmSpokePoolIndexerDataHandler implements IndexerDataHandler {
       message: `Found events for ${this.getDataIdentifier()}`,
       events: {
         depositEvents: events.depositEvents.length,
-<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


packages/indexer/src/data-indexing/service/Indexer.ts
```diff
@@ -185,4 +184,42 @@ export class Indexer {
       isBackfilling,
     };
   }
+
+  protected async getLatestBlockNumber(): Promise<number> {
```
melisaguevara, line 188: I kept the name as `getLatestBlockNumber` but we can improve this. I just couldn't think of a better option for now.

daywiss (approved):

PR #225, opened 2025-03-25
fix: reduce initial update time of hubpool and config clients

amateima:
This PR speeds up the hubpool and config clients update() function by leveraging correctly the Redis cache. During my local testing, the initial update from deployment block was reduced to ~25 seconds down from 200+ seconds

### DIFF
--- a/packages/indexer/src/web3/constants.ts
+++ b/packages/indexer/src/web3/constants.ts
@@ -26,12 +26,11 @@ const DEFAULT_NO_TTL_DISTANCE: { [chainId: number]: number } = {
 };
 
 export function getNoTtlBlockDistance(chainId: number) {
-  return 1_000_000_000_000;
-  // const noTtlBlockDistance = DEFAULT_NO_TTL_DISTANCE[chainId];
-  // if (!noTtlBlockDistance) {
-  //   throw new Error(`No noTtlBlockDistance found for chainId: ${chainId}`);
-  // }
-  // return noTtlBlockDistance;
+  const noTtlBlockDistance = DEFAULT_NO_TTL_DISTANCE[chainId];
+  if (!noTtlBlockDistance) {
+    throw new Error(`No noTtlBlockDistance found for chainId: ${chainId}`);
+  }
+  return noTtlBlockDistance;
 }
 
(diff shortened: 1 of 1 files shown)


packages/indexer/src/redis/redisCache.ts
```diff
@@ -13,7 +13,7 @@ export class RedisCache implements across.interfaces.CachingMechanismInterface {
     value: ObjectType,
     ttl?: number,
   ): Promise<string | undefined> {
-    if (ttl !== undefined) {
+    if (ttl !== undefined && ttl !== Number.POSITIVE_INFINITY) {
```
amateima, line 16: I saw the SDK setting `TTL` to Infinity which from my testing is an invalid value. Until it's fixed upstream in the SDK, I handle this case explicitly here

packages/indexer/src/web3/constants.ts
```diff
@@ -26,12 +26,11 @@ const DEFAULT_NO_TTL_DISTANCE: { [chainId: number]: number } = {
...
-  // const noTtlBlockDistance = DEFAULT_NO_TTL_DISTANCE[chainId];
-  // if (!noTtlBlockDistance) {
-  //   throw new Error(`No noTtlBlockDistance found for chainId: ${chainId}`);
-  // }
-  // return noTtlBlockDistance;
+  const noTtlBlockDistance = DEFAULT_NO_TTL_DISTANCE[chainId];
+  if (!noTtlBlockDistance) {
+    throw new Error(`No noTtlBlockDistance found for chainId: ${chainId}`);
+  }
+  return noTtlBlockDistance;
```
amateima, line 33: Enable no ttl block distance again, now that the Redis instance is evicting the least recently used keys.

packages/indexer/src/redis/redisCache.ts
melisaguevara: Yes, I also found this. I think it is already part of the stage branch.

amateima: Seems like so, yes! I was not aware of it

daywiss (approved):

melisaguevara (approved):

PR #226, opened 2025-03-25
feat: add filled deposits endpoint

daywiss:
# motivation
Add endpoint for dart to query recently filled deposits

# changes
This adds separate endpoint at deposits/filled, which returns deposit data + fill data for filled deposits only

This logic was not combined with unfilled because the data needed is slightly different.  i think its possible to combine all these calls but may make it a more complex than necessary

### DIFF
--- a/packages/indexer-api/src/dtos/deposits.dto.ts
+++ b/packages/indexer-api/src/dtos/deposits.dto.ts
@@ -21,7 +21,11 @@ export const DepositsParams = s.object({
   // some kind of pagination options, skip could be the start point
   skip: s.optional(stringToInt),
   // pagination limit, how many to return after the start, note we convert string to number
-  limit: s.defaulted(stringToInt, 50),
+  limit: s.refine(
+    s.defaulted(stringToInt, 50),
+    "maxLimit",
+    (value) => value <= 1000 || "Limit must not exceed 1000",
+  ),
 });
 
 export type DepositsParams = s.Infer<typeof DepositsParams>;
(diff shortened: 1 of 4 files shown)


packages/indexer-api/src/dtos/deposits.dto.ts
```diff
@@ -40,14 +40,14 @@ export const DepositParams = s.object({
...
 
-export const UnfilledDepositsParams = s.object({
+export const FilterDepositsParams = s.object({
   originChainId: s.optional(stringToInt),
   destinationChainId: s.optional(stringToInt),
   startTimestamp: s.optional(stringToInt),
   endTimestamp: s.optional(stringToInt),
-  minPendingSeconds: s.optional(stringToInt),
   skip: s.defaulted(stringToInt, 0),
   limit: s.defaulted(stringToInt, 50),
```
amateima, line 49: Can we provide a max value too? Like 100

daywiss: hmm yea potentially, let me loook into it

melisaguevara (commented):
looks good! Only one comment

packages/indexer-api/src/services/deposits.ts
```diff
@@ -3,13 +3,18 @@ import { DataSource, entities } from "@repo/indexer-database";
...
 import {
   DepositNotFoundException,
   IncorrectQueryParamsException,
   IndexParamOutOfRangeException,
 } from "./exceptions";
+import {
+  FilledV3Relay,
+  RelayHashInfo,
+  V3FundsDeposited,
+} from "../../../indexer-database/dist/src/entities";
```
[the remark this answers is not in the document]
daywiss: good catch, let me fix that

packages/indexer-api/src/dtos/deposits.dto.ts
daywiss: i added a validation check to max 1000. seems like 100 is a bit small

amateima (approved):

melisaguevara (approved):

packages/indexer-api/src/services/deposits.ts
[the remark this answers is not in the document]
melisaguevara: We can delete these lines 👍🏼

PR #228, opened 2025-03-25
chore: add transaction hash to deposits unique constraint

melisaguevara:
To handle Solana reorgs, we’ve added `transactionHash` to the unique constraint of the deposits table, now defined as (relayHash, blockNumber, transactionHash, logIndex).

Unlike EVM chains, Solana lacks a global log index per block, so we can’t rely on logIndex alone to uniquely identify transactions. Including transactionHash ensures uniqueness, as it’s generated from transaction data and remains consistent during reorgs, even if the blockNumber (or slot) changes.

### DIFF
--- a/packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
+++ b/packages/indexer-database/src/entities/evm/V3FundsDeposited.ts
@@ -9,9 +9,10 @@ import {
 } from "typeorm";
 
 @Entity({ schema: "evm" })
-@Unique("UK_v3FundsDeposited_relayHash_block_logIdx", [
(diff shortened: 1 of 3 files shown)


daywiss (approved):

james-a-morris:
> Unlike EVM chains, Solana lacks a global log index per block, so we can’t rely on logIndex alone to uniquely identify transactions.

To clarify on this - each signature has an ordering of events. And, the only transactions we're referencing are transactions of slots that have already been confirmed

james-a-morris (commented):
https://github.com/across-protocol/indexer/pull/228#issuecomment-2752547655

james-a-morris (approved):

amateima (approved):

PR #229, opened 2025-03-25
improve(priceWorker): enable price worker with env

melisaguevara:
This PR adds a new env var that let us enable or disable the price worker. It is useful when you work locally and want to avoid price messages to be processed.


packages/indexer/src/parseEnv.ts
```diff
@@ -200,6 +201,9 @@ export function envToConfig(env: Env): Config {
   const enableBundleBuilder = env.ENABLE_BUNDLE_BUILDER
     ? env.ENABLE_BUNDLE_BUILDER === "true"
     : true;
+  const enablePriceWorker = env.ENABLE_PRICE_WORKER
+    ? env.ENABLE_PRICE_WORKER === "true"
+    : false;
```
amateima, line 206: I think we should keep the default behaviour to be `enabled`

amateima (approved):

daywiss (approved):

james-a-morris (approved):

PR #234, opened 2025-03-26
fix: remove column from pool rebalance save

amateima:


melisaguevara (approved):

amateima:
> To confirm which column will this affect?

l2Token

PR #238, opened 2025-03-26
feat(api): add unmatched fills endpoint

daywiss:
# motivation
add unmatched fills endpoint

# changes
adds a new route /fills/unmatched

# bug fixes
"skip" is not working right for me, didnt work until i used "offset". plz look to see if that make sense

### DIFF
--- a/packages/indexer-api/src/services/deposits.ts
+++ b/packages/indexer-api/src/services/deposits.ts
@@ -92,7 +92,7 @@ export class DepositsService {
     }
 
     if (params.skip) {
-      queryBuilder.skip(params.skip);
+      queryBuilder.offset(params.skip);
     }
 
     if (params.limit) {
@@ -244,7 +244,7 @@ export class DepositsService {
       );
     }
 
-    queryBuilder.skip(skip);
+    queryBuilder.offset(skip);
     queryBuilder.limit(limit);
 
     return queryBuilder.execute();
@@ -311,7 +311,7 @@ export class DepositsService {
       );
     }
 
-    queryBuilder.skip(skip);
+    queryBuilder.offset(skip);
     queryBuilder.limit(limit);
 
     return queryBuilder.execute();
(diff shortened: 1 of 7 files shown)


packages/indexer-api/src/services/deposits.ts
```diff
@@ -92,7 +92,7 @@ export class DepositsService {
     }
 
     if (params.skip) {
-      queryBuilder.skip(params.skip);
+      queryBuilder.offset(params.skip);
```
melisaguevara, line 95: Why are we changing this? I think we should continue to use skip. From the docs, about offset:

> Sets OFFSET - selection offset. NOTE that it may not work as you expect if you are using joins. If you want to implement pagination, and you are having join in your query, then use the skip method instead.

melisaguevara (approved):
looks good! Only one question

packages/indexer-api/src/services/deposits.ts
daywiss: let me double check this

daywiss: using skip in the code, this is what i see:
```
// skip 0
curl "localhost:8080/deposits?limit=1&skip=0" | jq
//returns
[
  {
    "id": 226221,
     // ...
  }
]

// skip 100
curl "localhost:8080/deposits?limit=1&skip=100" | jq
// returns
[
  {
    "id": 226221,
     //  ...
  }
]
```
when querying the api, skip does not change the returned value

when using offset heres the result:
```
// skip 0
curl "localhost:8080/deposits?limit=1&skip=0" | jq
//returns
[
  {
    "id": 226221,
     // ...
  }
]

// skip 100
curl "localhost:8080/deposits?limit=1&skip=100" | jq
[
  {
    "id": 226121,
     // ...
  }
]
```

you can see here that the id is 100 lower than the first one, as expected

daywiss: @melisaguevara it seems like since you checked behavior, this is isolated to my setup somehow.  will revert this back to skip from limit

PR #239, opened 2025-03-26
feat: support Lens

melisaguevara:
This PR:
- Adds Lens constants
- Adds support for custom headers as we need that for the only Lens provider we have right now. The approach for this is similar to the one in the bots (see [here](https://github.com/across-protocol/relayer/pull/2152)) with the difference that there's no way to specify the provider it should apply to since we don't have named providers in the indexer.


packages/indexer/src/web3/RetryProvidersFactory.ts
```diff
@@ -34,6 +35,10 @@ export class RetryProvidersFactory {
         throw new Error(`Invalid provider urls found for chainId: ${chainId}`);
       }
       const standardTtlBlockDistance = getChainCacheFollowDistance(chainId);
+      // TODO: Headers are currently applied globally to all providers for a chain.
+      // Need to refactor to support provider-specific headers, as only some providers
+      // (e.g., Lens) require custom headers.
+      const headers = parseProviderHeaders(chainId);
```
melisaguevara, line 41: lmk if you all think this is ok for now or if you think we should improve it by using named providers or using other approach.

amateima (approved):

PR #242, opened 2025-03-27
fix: lens block time

melisaguevara:


packages/indexer/src/web3/constants.ts
```diff
@@ -117,7 +117,7 @@ export const BLOCK_TIME_SECONDS: { [chainId: number]: number } = {
   [CHAIN_IDs.BLAST]: 2,
   [CHAIN_IDs.BOBA]: 2,
   [CHAIN_IDs.INK]: 1,
-  [CHAIN_IDs.LENS]: 1,
+  [CHAIN_IDs.LENS]: 600, // TODO: Update to 1 when Lens go live
```
melisaguevara, line 120: Since they are not live yet they are not producing blocks at this rate.

james-a-morris (approved):

PR #254, opened 2025-04-03
feat: add speedup array to deposits

daywiss:
# motivation
frontend needs speedup info on deposits

# changes
This adds a secondary query to get all speedups related to the deposits returned.

### DIFF
--- a/packages/indexer-api/src/dtos/deposits.dto.ts
+++ b/packages/indexer-api/src/dtos/deposits.dto.ts
@@ -109,6 +109,7 @@ export type DepositReturnType = {
   speedups?: {
     transactionHash: string;
     updatedRecipient: string;
+    updatedOutputAmount: string;
     updatedMessage: string;
     blockNumber: number;
   }[];
--- a/packages/indexer-api/src/dtos/deposits.dto.ts
+++ b/packages/indexer-api/src/dtos/deposits.dto.ts
@@ -106,10 +106,10 @@ export type DepositReturnType = {
   swapToken?: string;
   swapTokenAmount?: string;
 
-  speedups?: Array<{
+  speedups?: {
     transactionHash: string;
     updatedRecipient: string;
     updatedMessage: string;
     blockNumber: number;
-  }>;
+  }[];
 };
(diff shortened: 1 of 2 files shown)


james-a-morris (approved):

packages/indexer-api/src/dtos/deposits.dto.ts
```diff
@@ -105,4 +105,11 @@ export type DepositReturnType = {
   swapTransactionHash?: string;
   swapToken?: string;
   swapTokenAmount?: string;
+
+  speedups?: Array<{
```
james-a-morris, line 109: OOC why not `{ }[]` vs `Array< >`

packages/indexer-api/src/services/deposits.ts
```diff
@@ -347,7 +405,36 @@ export class DepositsService {
...
+          "speedup.updatedRecipient as updatedRecipient",
+          "speedup.updatedMessage as updatedMessage",
+          "speedup.blockNumber as blockNumber",
+        ])
+        .getRawMany();
+
+      deposit.speedups = speedups;
+    });
+
+    await Promise.all(speedupPromises);
```
james-a-morris, line 435: Normal convention is to nest this

james-a-morris, line 435: I'd recommend we transform this so we return the deposit in the mapping with the speedups added so that `deposits` isn't mutated

packages/indexer-api/src/services/deposits.ts
```diff
@@ -143,7 +143,36 @@ export class DepositsService {
...
+          {
+            depositId: deposit.depositId,
+            originChainId: deposit.originChainId,
+          },
+        )
+        .select([
+          "speedup.transactionHash as transactionHash",
+          "speedup.updatedRecipient as updatedRecipient",
+          "speedup.updatedMessage as updatedMessage",
+          "speedup.blockNumber as blockNumber",
```
melisaguevara, line 166: should we also include `updatedOutputAmount`?

packages/indexer-api/src/dtos/deposits.dto.ts
daywiss: hmm yea i dont think it matters, i think we typically do {}[]

packages/indexer-api/src/services/deposits.ts
daywiss: good catch

packages/indexer-api/src/services/deposits.ts
daywiss: i think i get what u are saying, will update

amateima (approved):

PR #270, opened 2025-04-24
improve: rename return fields to match deposit status

daywiss:
# motivation
we want to see return types the same as the deposit status return type

# changes
Only a couple fields needed to change in the other endpoints


amateima (approved):
Approved, but this is a breaking change for the FE. I need to prepare the Transactions Page for the new fields names

PR #295, opened 2025-05-15
fix: normalize addresses for deposits endpoint

melisaguevara:
PancakeSwap was having issues when querying deposits because the endpoint is case sensitive.
To fix it, we can use ethers utility of getAddress which will parse the address and return it in the way we expect it.


amateima (commented):
I'm wondering if there's something in the sdk that can be used to get the address. I would avoid importing a different ethers version because of potential dependencies conflicts in the future

melisaguevara:
> I'm wondering if there's something in the sdk that can be used to get the address. I would avoid importing a different ethers version because of potential dependencies conflicts in the future

@amateima fair call! I've replaced it with an util function from the sdk.

PR #299, opened 2025-05-19
feat: use SVM spoke pool client to query events

melisaguevara:
This PR is part of the required changes to support Solana in the indexer. It replaces the SVMEventsClient with the SVMSpokePoolClient to query SVM events.

Upcoming PRs will store events in the database and process them using the SpokePoolProcessor.

### DIFF
--- a/packages/indexer/src/data-indexing/service/SvmSpokePoolIndexerDataHandler.ts
+++ b/packages/indexer/src/data-indexing/service/SvmSpokePoolIndexerDataHandler.ts
@@ -169,7 +169,7 @@ export class SvmSpokePoolIndexerDataHandler implements IndexerDataHandler {
     const endTimeToGetSlotTimes = performance.now();
 
     this.logger.debug({
-      at: "SpokePoolIndexerDataHandler#fetchEventsByRange",
+      at: "SvmSpokePoolIndexerDataHandler#fetchEventsByRange",
       message: "Time to update protocol clients",
       timeToUpdateProtocolClients: timeToUpdateProtocolClients - initialTime,
       timeToUpdateSpokePoolClient:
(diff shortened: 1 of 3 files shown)


packages/indexer/src/data-indexing/service/SvmSpokePoolIndexerDataHandler.ts
```diff
@@ -114,56 +115,112 @@ export class SvmSpokePoolIndexerDataHandler implements IndexerDataHandler {
...
+      ...fundsDepositedEvents.map((deposit) => deposit.blockNumber),
+      ...filledRelayEvents.map((fill) => fill.blockNumber),
+    ];
+
+    const startTimeToGetSlotTimes = performance.now();
+    const slotTimes = await this.getSlotTimes(slots);
+    const endTimeToGetSlotTimes = performance.now();
+
+    this.logger.debug({
+      at: "SpokePoolIndexerDataHandler#fetchEventsByRange",
```
melisaguevara, line 172:
```suggestion
      at: "SvmSpokePoolIndexerDataHandler#fetchEventsByRange",
```

melisaguevara:
> OOC - now that we've abstracted the spoke client - do we need a separate data indexer?

@james-a-morris Since there are a couple things that will continue to be different, I think it makes sense to have a new class for this indexer. I also think I prefer it over having multiple if clauses in a single class. This indexer would also be responsible for formatting data so that we are able to have a single instance of both the `SpokePoolRepository` and the `SpokePoolProcessor`.

Some of the things that are different:
- We'll have to format some addresses from hexadecimal back to base58
- Getting slot times differs from getting block times.
- Providers are different, if we keep a single class we'll need to type check the provider each time we use it.
- We don't need swap events data for SVM.

We could for sure refactor this a bit so that there's not too much duplicated code. I could work on it once we have the whole indexing logic ready.

@james-a-morris @amateima let me know what you think.

amateima (approved):

PR #302, opened 2025-05-20
chore: update hubpool events chain columns to support solana chainId

melisaguevara:
This PR changes the column type of `root_bundle_executed.chainId` and `set_pool_rebalance_route.destinationChainId` from `int` to `bigint`. This change is required because Solana's chain id is greater than the max allowed value in postgres int columns.

### DIFF
--- a/packages/indexer-database/src/entities/evm/RootBundleExecuted.ts
+++ b/packages/indexer-database/src/entities/evm/RootBundleExecuted.ts
@@ -23,7 +23,7 @@ export class RootBundleExecuted {
   @Column()
   groupIndex: number;
 
-  @Column()
+  @Column({ type: "bigint" })
   chainId: number;
 
   @Column({ type: "jsonb" })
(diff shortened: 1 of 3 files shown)


packages/indexer-database/src/entities/evm/RootBundleExecuted.ts
```diff
@@ -23,7 +23,7 @@ export class RootBundleExecuted {
   @Column()
   groupIndex: number;
 
-  @Column()
+  @Column({ type: "bigint" })
   chainId: number;
```
amateima, line 27: shouldn't the type be changed to string? As far as I know, `bigint` is > than the largest JS safe integer

melisaguevara: Yes it is, but since the SDK parses chain IDs into Number, we likely won’t need to store any value larger than JS’s maximum safe integer.

amateima: Indeeed, but when the events will be queried, TypeORM will cast the field to string :(

amateima (approved):

PR #303, opened 2025-05-20
chore: update spoke pool events chain id column types I

melisaguevara:
This PR updates the chainId columns from int to bigint for: `v3FundsDeposited`, `filledV3Relay` and `requestedSlowFill` tables. This is required to support Solana's chain id.


packages/indexer-database/src/entities/evm/FilledV3Relay.ts
```diff
@@ -26,10 +26,10 @@ export class FilledV3Relay {
   @Column({ type: "decimal" })
   depositId: string;
 
-  @Column()
+  @Column({ type: "bigint" })
   originChainId: number;
```
amateima, line 30: Same comment as in the PR for the HubPool events

PR #304, opened 2025-05-20
chore: update spoke pool events chain id columns II

melisaguevara:
This PR updates the 'chainId' column for `ExecutedRelayerRefundRoot`, `RelayedRootBundle` and `TokensBridged` tables. This is required to support Solana's chain id.


packages/indexer-database/src/entities/evm/ExecutedRelayerRefundRoot.ts
```diff
@@ -16,7 +16,7 @@ export class ExecutedRelayerRefundRoot {
   @PrimaryGeneratedColumn()
   id: number;
 
-  @Column()
+  @Column({ type: "bigint" })
   chainId: number;
```
amateima, line 20: Same comment as in the PR for the HubPool events

amateima (approved):

PR #307, opened 2025-05-21
feat: support Solana

melisaguevara:
This PR prepares the indexer to support Solana by:
- Fetching SpokePool events.
- Storing them in the database formatting addreses back from bytes32 to base58 when needed.
- Processing events using the existing SpokePoolProcessor.
- Fetching fill gas amounts and token prices.
- Adapting the API for it to support EVM and SVM addresses.


packages/indexer-database/src/migrations/1747857452570-nullableCaller.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "evm"."tokens_bridged" ALTER COLUMN "caller" SET NOT NULL`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "evm"."executed_relayer_refund_root" ALTER COLUMN "caller" SET NOT NULL`,
+    );
+  }
+}
```
melisaguevara, line 23: This field is not available for svm events.
We could add it to the sdk but since there's no clear usage for it I think it's ok to avoid it.

packages/indexer/src/data-indexing/service/SpokePoolIndexerDataHandler.ts
```diff
@@ -459,17 +459,13 @@ export class SpokePoolIndexerDataHandler implements IndexerDataHandler {
...
-      "TokensBridged",
+      "FilledRelay",
       "FundsDeposited",
-      "RequestedSpeedUpDeposit",
+      "RelayedRootBundle",
       "RequestedSlowFill",
-      "FilledRelay",
+      "RequestedSpeedUpDeposit",
+      "TokensBridged",
     ]);
```
melisaguevara, line 469: Updating this list since sdk new version dropped V3 support.

packages/indexer-database/src/migrations/1749067055693-requestSpeedUp.ts
```diff
@@ -0,0 +1,17 @@
...
+      `ALTER TABLE "evm"."requested_speed_up_v3_deposit" ALTER COLUMN "originChainId" TYPE bigint`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "evm"."requested_speed_up_v3_deposit" ALTER COLUMN "originChainId" TYPE integer`,
+    );
+  }
+}
```
melisaguevara, line 17: I thought this type change was not needed for this table since we don't have RequestedSpeedUp in Solana, but it was indeed needed since we query this table for the API endpoints and at times we'll use solana's chain id as input for the query.

amateima (approved):

PR #315, opened 2025-06-11
improve: refactor spoke pool factory to support svm

melisaguevara:
This PR refactors the SpokePoolClient factory for it to be able to instantiate evm or svm clients.

It also introduces a custom implementation of the SVMSpokePoolClient that lets us disable block look ups in the hub pool client, same as we do for EVM.

### DIFF
--- a/packages/indexer/src/services/BundleIncludedEventsService.ts
+++ b/packages/indexer/src/services/BundleIncludedEventsService.ts
@@ -235,36 +235,19 @@ export class BundleIncludedEventsService extends RepeatableTask {
           return [chainId, null];
         }
 
-        if (chainIsSvm) {
-          const spokePoolClient =
-            await across.clients.SVMSpokePoolClient.create(
-              this.logger,
-              this.hubPoolClient,
-              chainId,
-              BigInt(deployedBlockNumber),
-              {
-                from: startBlock,
-                to: cappedEndBlock,
-              },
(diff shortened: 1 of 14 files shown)


packages/indexer/src/utils/contractFactoryUtils.ts
```diff
@@ -34,6 +39,31 @@ abstract class ContractClientFactory<
   ): ClientType;
 }
 
+abstract class AsyncContractClientFactory<
```
melisaguevara, line 42: Resolving constructor params for both the SVMSpokePoolClient and the SVMCpiEventsClient requires asynchronous calls (for example to resolve some PDAs), then we need this factory to handle async instantiations.

packages/indexer/src/services/BundleIncludedEventsService.ts
```diff
@@ -235,36 +235,19 @@ export class BundleIncludedEventsService extends RepeatableTask {
...
-              startBlock,
-              cappedEndBlock,
-              {
-                hubPoolClient: this.hubPoolClient,
-              },
-              false,
-            ),
-          ];
-        }
+        const enableCaching = chainIsSvm ? true : false;
```
melisaguevara, line 238: @amateima I think for SVM we should keep caching enabled. Since querying events there will always start from the most recent slot, whenever we query an old block range it might take too long. We could off course evaluate how this performs and disable caching later if required. Let me know what you think.

PR #317, opened 2025-06-12
feat: support new Address class from SDK

melisaguevara:
This PR prepares the indexer to support the new Address class from the SDK.


packages/indexer/package.json
```diff
@@ -21,9 +21,9 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@across-protocol/constants": "^3.1.67",
-    "@across-protocol/contracts": "^4.0.12",
-    "@across-protocol/sdk": "^4.2.15",
+    "@across-protocol/constants": "^3.1.68",
+    "@across-protocol/contracts": "^4.0.13",
```
melisaguevara, line 25: I also updated contracts but this version is still pointing to old Solana programs. We'll have to update once again when new version is published.

amateima (approved):

PR #343, opened 2025-08-12
chore: fix refunded status implementation

amateima:


packages/indexer/src/services/RefundedDepositsStatusService.ts
```diff
@@ -0,0 +1,145 @@
...
+        "be.relayHash = dep.internalHash AND be.eventChainId = dep.originChainId AND be.eventBlockNumber = dep.blockNumber AND be.eventLogIndex = dep.logIndex",
+      )
+      .innerJoin(entities.RelayHashInfo, "rhi", "dep.id = rhi.depositEventId")
+      .where("be.type = :expiredDeposit", {
+        expiredDeposit: entities.BundleEventType.ExpiredDeposit,
+      })
+      .andWhere("dep.originChainId = :chainId", { chainId })
+      .andWhere("rhi.status = :status", {
+        status: entities.RelayStatus.Expired,
+      })
```
amateima, line 43: join with `RelayHashInfo` and filter only the expired deposits

packages/indexer/src/services/RefundedDepositsStatusService.ts
```diff
@@ -0,0 +1,145 @@
...
+            .update()
+            .set({
+              status: entities.RelayStatus.Refunded,
+              depositRefundTxHash:
+                executedRelayerRefundRootEvent.transactionHash,
+            })
+            .where("id = :rowToUpdateId", {
+              rowToUpdateId: rowToUpdate.id,
+            })
+            .returning("*")
```
melisaguevara, line 131: nit: I think we are not doing anything with this data after executing this function, right? Should we return only id maybe?

packages/indexer/src/services/RefundedDepositsStatusService.ts
```diff
@@ -0,0 +1,145 @@
...
+      entities.BundleEvent,
+    );
+    const refundEventsQb = bundleEventsRepository
+      .createQueryBuilder("be")
+      .innerJoinAndSelect("be.bundle", "bundle")
+      .innerJoinAndMapOne(
+        "be.deposit",
+        entities.V3FundsDeposited,
+        "dep",
+        "be.relayHash = dep.internalHash AND be.eventChainId = dep.originChainId AND be.eventBlockNumber = dep.blockNumber AND be.eventLogIndex = dep.logIndex",
```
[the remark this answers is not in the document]
amateima: I think a schema change is out of the scope of this PR

packages/indexer/src/services/RefundedDepositsStatusService.ts
amateima: Good point. Edited 👍

PR #368, opened 2025-09-16
feat: enable HyperEVM

ashwinrava:


packages/indexer/src/data-indexing/service/constants.ts
```diff
@@ -9,6 +9,7 @@ const finalisedBlockBufferDistances: Record<number, number> = {
   [CHAIN_IDs.BASE]: 60,
   [CHAIN_IDs.BLAST]: 60,
   [CHAIN_IDs.BSC]: 60,
+  [CHAIN_IDs.HYPEREVM]: 60,
```
ashwinrava, line 12: Couldn't find reference values in [bot configs](https://github.com/UMAprotocol/bot-configs/blob/master/serverless-bots/across-v2-bot-config.json). I think they used to be there in an older commit as the comment says above.

PR #381, opened 2025-09-25
feat: index target chain action events

ashwinrava:
- Added new DB column to `RelayHashInfo` called `actionsTargetChainId`
- Added concept of "target chain actions" which indicates scenarios where there is an intermediary chain like HyperEVM
- Parse transaction receipts for HyperEVM -> HyperCore transfer and match with fill if transfer amount = fill amount
- Update `RelayHashInfo.actionsTargetChainId` with 1337 for HyperCore

Initial tests are working as expected:

- Optimism to HyperEVM: https://hyperevmscan.io/tx/0xcebd9faf1b18648476712218b76d66bae45a7b86943d5df7f82b8aa7262ddbee
- Destination transfer to HyperCore detected in local indexer run and new DB column updated

Screenshot 2025-09-25 at 9 32 20 PM

Next steps:

- Update UI


packages/indexer/src/utils/terminalTransferUtils.ts
```diff
@@ -0,0 +1,92 @@
...
+        if (terminalTransferChainId) {
+          acc[fill.transactionHash] = {
+            fills: fills.filter(
+              (f) =>
+                f.transactionHash.toLowerCase() ===
+                fill.transactionHash.toLowerCase(),
+            ),
+            terminalTransferChainId,
+          };
+        }
```
amateima, line 62: It is possible to have multiple fills to HyperEVM in the same txn, right? If so, what if all fills have the same amount, but only one transfer goes to HyperCore? Does the algorithm work in this case? Because you'd want to match only one fill to a terminal transfer event

melisaguevara (commented):
Looks good! Left some comments.

packages/indexer/src/utils/terminalTransferUtils.ts
ashwinrava: Good point, I added this safeguard in my latest commit 3acb431dfb4b41fb76e7d854bfaa59f0da1bf2df. Followed the same pattern that we do for deposit + swap pairs and fills + failed calls pairs.

packages/indexer/src/utils/terminalTransferUtils.ts
```diff
@@ -0,0 +1,92 @@
...
+          ? TERMINAL_TRANSFER_ADDRESSES[terminalAddress]
+          : null;
+
+        if (terminalTransferChainId) {
+          acc[fill.transactionHash] = {
+            fills: fills.filter(
+              (f) =>
+                f.transactionHash.toLowerCase() ===
+                fill.transactionHash.toLowerCase(),
+            ),
```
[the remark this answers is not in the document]
ashwinrava: Yeah true, removed in latest commit.

melisaguevara (approved):
Looks good! 😄

